PR #83, opened 2026-07-04
security: couple face templates to the enrolling camera (Plan 02)

### LINKED ISSUE #64: Face templates not bound to enrolling camera — attacker-supplied camera feeds spoof frames
## Finding

Face templates are not bound to the camera that enrolled them. An enrolled template could be matched against frames from *any* camera the daemon selects, with no check that the live capture device is the same (or same model) as the device used at enrollment.

## Impact / attack scenario

An attacker with physical access can plug in their own camera — a commodity or different-model USB device, or a programmable USB camera — and feed attacker-controlled or spoof frames into the auth path. Because templates were device-agnostic, a match on attacker-supplied frames was accepted the same as a match from the user's real (e.g. IR) camera. This is the "attacker plugs in their own camera to feed a spoof" vector, and it is a more robust bypass than defeating the IR heuristic alone.

Note the honest scope: VID:PID fingerprinting is model-granularity and a programmable USB device can forge it, so device coupling is advisory defense-in-depth, not attested root of trust.

## Where it lives

- `crates/facelock-camera/src/quirks.rs` / `device.rs` — device fingerprint (VID:PID / serial / by-path)
- `crates/facelock-store/src/migrations.rs`, `db.rs` — templates stored with no device identity column
- `crates/facelock-daemon/src/handler.rs`, `auth.rs` — compare loop does not filter by device

## How it was found

Internal security review, July 2026 (Plan 02).

tyvsmith:
## Summary

Binds each enrolled template to a fingerprint of the camera that captured it and, at auth, skips any template whose enrolling-camera fingerprint does not match the live camera (Plan 02). A swapped-in camera therefore degrades to the password fallback — it can never reach `PAM_SUCCESS`, and it never causes a lockout (fail SOFT). This is the robust successor to Plan 01's IR fix, closing the "attacker plugs in their own camera to feed a spoof" vector by identity rather than heuristic.

**Honest framing (documented):** the fingerprint is model-granularity (VID:PID) at best and forgeable by a programmable USB device — advisory defense-in-depth, NOT attestation.

- `facelock-core`: `DeviceFingerprint` + canonical `"vid:pid:serial"`, granularity matching, per-model gate, compare-set filter, and the Plan 04 AAD seam. `FaceModelInfo`/`ModelInfo` carry `device_id`.
- Config: `security.bind_templates_to_device` (default `true`), `device_match_granularity` (`"model"` default / `"unit"`), `bind_legacy_templates` (default `true`, allow-with-warn for legacy NULL rows).
- Camera: `device_fingerprint()` reads idVendor/idProduct/serial + v4l by-path from sysfs, tolerating all-missing (→ NULL, legacy-governed, never a lockout).
- Store: migration **V6** adds nullable `face_models.device_id` (mirrors V5); enroll threads the fingerprint; `list_models` reads it.
- Daemon + oneshot + direct auth loops filter the compare set by device and assert in code that a mismatch can never reach the success branch. `facelock list` shows each template's camera.

## Stacked base

This PR is **based on `sec/01-passive-antispoof-honesty`** (shared auth/camera path). It should be **retargeted to `main` after the parent (Plan 01) merges.** The diff shown against `sec/01` is the Plan 02 change only.

## Verification

- `cargo build` + `cargo test` (382 tests) + `cargo clippy -D warnings`.
- Live camera (real IR BRIO 046d:085e:849A5B02): `just test-arch-oneshot` 17/17 and `just test-arch-integration` 11/11. Container assertions: V6 applies on startup, `device_id` is non-null after enroll, a forged/overridden identity falls through (swap-in regression gate), legacy NULL rows still authenticate, and a pre-V6 DB migrates cleanly with no data loss.
- Unit tests cover fingerprint parsing, granularity matching, and the mismatch-never-succeeds assertion (camera-free core proof).

## Contract docs

`docs/contracts.md` and `docs/security.md` updated: schema **V6** (`device_id`), new config keys, new auth semantics, and the "advisory / spoofable by programmable USB" caveat (§1.D).

Closes #64

### DIFF
--- a/config/facelock.toml
+++ b/config/facelock.toml
@@ -209,6 +209,27 @@
 # Default: 3
 # min_auth_frames = 3
 
+# Device coupling — bind each enrolled template to the camera that captured it.
+# At auth, a template whose enrolling-camera fingerprint does not match the live
+# camera is skipped, so a swapped-in camera degrades to the password fallback
+# (never a match, never a lockout). Advisory defense-in-depth only: the
+# fingerprint is model-granularity (VID:PID) and forgeable by a programmable USB
+# device — NOT attestation. See docs/security.md §1.D.
+#
+# Default: true
+# bind_templates_to_device = true
+#
+# Match granularity: "model" (VID:PID, default — blocks a cross-model swap) or
+# "unit" (VID:PID:serial — also blocks a same-model swap, but requires a camera
+# that exposes a stable serial; enrollment at "unit" is refused otherwise).
+# device_match_granularity = "model"
+#
+# Allow legacy templates (enrolled before device coupling, or on a camera with
+# no readable USB identity — NULL device_id) to authenticate. Default true
+# (allow-with-warn) so upgrades never lock anyone out; set false to require every
+# template to carry a matching device id.
+# bind_legacy_templates = true
+
 # Rate limiting — prevents brute-force attempts.
 #
 # [security.rate_limit]
--- a/crates/facelock-camera/src/lib.rs
+++ b/crates/facelock-camera/src/lib.rs
@@ -11,4 +11,4 @@ pub use device::{
 };
 pub use ir_emitter::EmitterXuInfo;
 pub use preprocess::{check_ir_texture, clahe, extract_bbox_region, rgb_to_gray, yuyv_to_rgb};
-pub use quirks::{Quirk, QuirksDb};
+pub use quirks::{Quirk, QuirksDb, device_fingerprint};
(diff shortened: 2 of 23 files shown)


tyvsmith:
Landed in `main` via the security integration merge — sec/02-device-coupling's tip `80115d9` is an ancestor of `main` (`8a1418d`). GitHub can't auto-mark this MERGED because the PR's base was the stacked parent (sec/01), not `main`; the commits are fully in `main`. Closing as merged. Resolves #64.

Issue #70, opened 2026-07-04
TPM PCR binding off by default with no re-seal recovery path


tyvsmith:
## Finding

TPM PCR binding is off by default and there was no re-seal recovery path. Even once PCR enforcement works, `pcr_binding` defaulted to `false`, and there was no supported command to re-seal the key after a legitimate platform change (kernel/firmware update that alters a bound PCR).

## Impact / attack scenario

Two-sided problem. With PCR binding off by default, the default TPM mode only protects against extracting the sealed blob without the TPM — it does not bind the key to platform integrity, so users who assume TPM = measured-boot protection get less than they expect. But turning it *on* without a recovery path is also unsafe: a routine kernel/firmware update would change the bound PCR and permanently prevent unseal, effectively bricking the biometric key (password fallback still works, but the enrolled templates become unusable). The default flip must wait until an explicit `facelock reseal` recovery UX exists.

## Where it lives

- `crates/facelock-core/src/config.rs` — `pcr_binding` / `pcr_indices` defaults
- `crates/facelock-cli/src/commands/` — missing `reseal` recovery command

## How it was found

Internal security review, July 2026 (Plan 04, finding #12).

Issue #10, opened 2026-03-21
Install always shows setup instructions even when already configured


tyvsmith:
## Problem

`just install` always prints the same post-install message regardless of system state:

```
Installed. Two steps remaining:
  1. sudo facelock setup       (download face recognition models)
  2. sudo facelock enroll      (register your face)
```

This is misleading when the system is already fully configured with models, PAM, and enrollments.

Additionally, the uninstall only cleans PAM from `/etc/pam.d/sudo` using exact line matching, missing polkit-1 and hyprlock.

## Proposed Solution

**Install**: Check what's actually missing and show conditional output:
- Models present? (`ls /var/lib/facelock/models/*.onnx`)
- Config present? (`/etc/facelock/config.toml`)
- PAM configured? (`grep pam_facelock /etc/pam.d/sudo`)
- ORT installed? (check known library paths)
- Show "Installed and up to date." when everything is in place

**Uninstall**: Remove PAM lines from all known services (sudo, polkit-1, hyprlock) using module-name matching instead of exact string match.

Also adds omarchy-enable/disable justfile targets for hyprlock integration.

## Acceptance Criteria

- [ ] Fresh install shows relevant setup instructions
- [ ] Fully configured install shows "Installed and up to date."
- [ ] Missing ORT shows install instructions
- [ ] Uninstall cleans pam_facelock from sudo, polkit-1, and hyprlock

PR #85, opened 2026-07-04
security: encrypt-by-default, keyfile TOCTOU fix, enforced TPM PCR binding (Plan 04)

### LINKED ISSUE #67: Biometric templates stored in plaintext by default
## Finding

Biometric templates are stored in plaintext by default. `EncryptionMethod` defaulted to `None`, so a fresh install wrote face embeddings to the SQLite store as raw, unencrypted vectors.

## Impact / attack scenario

Anyone who can read the database file (a backup, a stolen disk, a misconfigured permission, or a root-level compromise) obtains the raw biometric templates. Face embeddings are sensitive, non-revocable biometric data — unlike a password they cannot be rotated. Storing them in the clear by default means the most sensitive asset in the system has no at-rest protection unless the user knew to opt in.

## Where it lives

- `crates/facelock-core/src/config.rs` — `EncryptionMethod` default (`None`)
- `crates/facelock-store` — embedding storage

## How it was found

Internal security review, July 2026 (Plan 04, finding #8).

### LINKED ISSUE #68: Encryption key file creation TOCTOU — window where key is world-readable
## Finding

The encryption key file is created with a TOCTOU window in which the key is world-readable. `generate_key_file` / `seal_key_to_file` wrote the key with `fs::write` and only *afterward* called `set_permissions(0o600)`. Between the create and the chmod, the file exists at the process umask default (e.g. `0644`), and a "write atomically" comment described behavior that was not actually atomic.

## Impact / attack scenario

A local attacker racing the key-file creation can `open()` the key during the window it is world-readable and capture the AES key that protects the biometric templates, defeating the at-rest encryption. This is a classic create-then-chmod race; the fix is to create the file at mode `0600` in a single `open(2)` with `O_CREAT|O_EXCL`.

## Where it lives

- `crates/facelock-tpm/src/sealing.rs` — `generate_key_file`, `seal_key_to_file`
- `crates/facelock-core/src/fs_security.rs` — atomic create-at-mode helper

## How it was found

Internal security review, July 2026 (Plan 04, finding #11 / keyfile TOCTOU).

### LINKED ISSUE #69: TPM PCR policy computed but not enforced at unseal
## Finding

The TPM PCR policy is computed but **not enforced** at unseal. The code built a PCR policy digest (`build_pcr_policy_digest`) for PCR-bound sealed objects, but the unseal path used `execute_with_nullauth_session` and never started a real policy session or replayed `PolicyPCR` over the sealed selection before `TPM2_Unseal`. The objects were also sealed with `with_user_with_auth(true)`, allowing USER-role auth to bypass the policy entirely.

## Impact / attack scenario

The PCR binding was cosmetic: the key would unseal regardless of the actual platform PCR state, so the protection users believe they have ("the key only releases if the boot/firmware measurement matches") did not hold. An attacker who moves the sealed blob to a tampered/different platform state — or who alters a bound PCR — could still unseal, because the policy digest was computed and then never used to gate the release. A regression test that seals with PCR binding, alters a bound PCR, and expects unseal to *fail* would have passed (unsealed) before the fix, proving the bug.

## Where it lives

- `crates/facelock-tpm/src/sealing.rs` — seal attributes/policy, unseal path, policy digest

## How it was found

Internal security review, July 2026 (Plan 04, finding #5).

tyvsmith:
## Summary

Stops shipping plaintext biometric templates by default, closes the key-file creation race, and makes TPM PCR binding actually enforced (Plan 04).

- **Encrypt-by-default (#8):** `EncryptionMethod` default `None` → `Keyfile`. The keyfile is auto-generated at `0600` on first use. Plaintext (`method="none"`) is refused at enroll unless `security.allow_plaintext=true` (explicit opt-in + warning). Auth degrades to password on any decrypt failure — never a lockout. Fixed a latent regression this exposed: the oneshot `facelock auth` binary read embeddings as plaintext only and is now routed through the decryption-aware loader (removed the plaintext-only `auth::authenticate` footgun).
- **Keyfile TOCTOU (#9):** `generate_key_file` / `seal_key_to_file` now create the file at mode `0600` in one `open(2)` via `fs_security::create_truncate_file` (+ `sync_all`), closing the create-then-chmod window. Real concurrency test added.
- **TPM PCR enforcement (#10):** PCR-bound sealed objects use `userWithAuth=false` and a self-describing `0x03` blob format recording the PCR index list. The unseal path now starts a real policy session and replays `PolicyPCR` against current PCRs, so a changed bound PCR makes unseal fail. New read-only `facelock tpm unseal-check` diagnostic.
- **Opt-in hard device binding:** `security.bind_device_aad` folds `device_id` into the AES-GCM AAD so a template can't be decrypted under a different camera (default off — fails closed on unstable IDs). This is the Plan 02 AAD seam realized.

## PCR default (deferred)

The `facelock reseal` recovery command is added (re-seals the key under current PCRs, recovering from the sealed blob or the key backup), but `pcr_binding` stays default `false` — the default flip is deliberately deferred to a human decision now that the recovery path exists.

## Stacked base

This PR is **based on `sec/02-device-coupling`** (it consumes the `device_id` AAD seam). It should be **retargeted to `main` after the parent chain (Plans 01 → 02) merges.** The diff shown against `sec/02` is the Plan 04 change only.

## Verification

- `cargo build` / `cargo test` / `cargo clippy -D warnings` green; real create/stat concurrency test for the keyfile race.
- Camera hosts (`just test-arch-oneshot` / `test-arch-integration`): assertion that a fresh enroll's `face_embeddings` blob is NOT raw 2048-byte f32 (encrypted/sealed flag set) and that enroll→auth still round-trips.
- **TPM PCR regression (swtpm, no camera) — `just test-deb-tpm-pkg`:** seal with PCR binding → unseal OK; extend/alter a bound PCR → unseal **fails**; `facelock reseal` restores unseal. This proves finding #10 is actually fixed (the same test would have unsealed before the fix).

## Contract docs

`docs/contracts.md` and `docs/security.md` updated: default encryption method change, TPM unseal semantics, new `reseal` command, AAD binding, config defaults; config template documents the new knobs.

Closes #67
Closes #68
Closes #69
Addresses #70 (adds `facelock reseal` recovery path; default flip deliberately deferred)


tyvsmith:
Landed in `main` via the security integration merge — sec/04-storage-crypto-honesty's tip `15756a4` is an ancestor of `main` (`8a1418d`). Stacked base (sec/02) prevented GitHub's auto-merge detection; the commits are fully in `main`. Closing as merged. Resolves #67, #68, #69 (#70 intentionally left open — reseal recovery path added, default flip deferred).

Issue #89, opened 2026-08-07
fix: support Intel IPU7 cameras exposed through v4l2-relayd


felipecsl:
Disclaimer: This was written by my agent after troubleshooting facelock on my laptop.

## Summary

Facelock 0.1.4 could not enroll a face on a Dell XPS 14 DA14260 with an Intel IPU7 camera. The initial CLI error was generic:

```text
Enrollment failed: D-Bus Enroll call failed
```

The daemon received the call, but it selected the raw IPU7 capture node and captured zero usable frames. The system provides a processed camera through `v4l2-relayd` and `v4l2loopback` at `/dev/video50`, but that path also needs format and relay changes before Facelock can use it.

Enrollment and recognition work after the changes below.

## Environment

```text
Device: Dell XPS 14 DA14260
OS: Arch Linux / Omarchy
Kernel: 7.1.3-arch1-1-ptl
Facelock: 0.1.4
intel-ipu7-camera: 1.0.5-1
v4l2-relayd: 0.2.0
v4l2loopback-dkms: 0.15.4-1
GStreamer: 1.28.5
```

The physical sensor is `ov08x40-uf`. The IPU7 stack exposes raw capture nodes at `/dev/video0` through `/dev/video31` and a processed loopback camera at `/dev/video50`.

### 1. Auto-detection selected an unusable raw node

The daemon selected `/dev/video0`:

```text
auto-detected camera device device=/dev/video0 name=ipu7 ir=false
```

`media-ctl` showed that this node carries raw Bayer data:

```text
SGRBG10_1X10/3856x2176
```

Facelock supports `GREY`, `YUYV`, and `MJPG`, so it could not process this raw IPU7 node. The enrollment loop hid the repeated capture errors and ended with:

```text
insufficient face captures during enrollment captured=0 required=3
```

Setting the Facelock device explicitly fixed this part:

```toml
[device]
path = "/dev/video50"
```

It can be useful for auto-detection to exclude capture nodes that do not advertise a format Facelock can decode.

### 2. The processed camera used unsupported NV12

The default IPU7 relay exposed `/dev/video50` as:

```text
Pixel Format: NV12
```

Facelock 0.1.4 does not decode NV12. The relay had to convert its output to YUYV/YUY2:

```toml
VIDEOSRC="icamerasrc device-name=ov08x40-uf sharpness=80 ev=-1 saturation=10 ! videoflip method=rotate-180 ! videoconvert"
FORMAT=YUY2
[...]

PR #90, opened 2026-08-08
feat(camera): NV12/Y16 decode + skip undecodable devices in auto-detection

tyvsmith:
## Summary

- decode NV12 (semi-planar 4:2:0) and Y16 (16-bit IR grayscale, bit-depth aware)
- exclude devices with no decodable format from every auto-detection tier, so raw Bayer sensor nodes cannot be selected
- fail `Camera::open` with the advertised and supported format lists instead of negotiating a format every capture then rejects
- negotiate in priority order `quirk format_preference > GREY > Y16 > YUYV > NV12 > MJPG`
- pin the Y16 16-bit-to-8-bit shift once per camera session, from a quirk's `y16_bit_depth` when declared and a burst of frames at open otherwise
- reject a padded `bytesperline` at open rather than decoding sheared frames
- document the Intel IPU6/IPU7 + v4l2-relayd path, and why `require_ir` cannot be satisfied on that hardware

## Why

Auto-detection selected the raw IPU7 capture node on a Dell XPS 14, negotiated a format it could not decode, and captured zero usable frames — surfacing only as "insufficient face captures during enrollment" (#89). The undecodable-device exclusion is the actual fix for that class of silent failure and is independent of any one camera. NV12 support additionally makes the processed `v4l2-relayd` loopback usable without reconfiguring the relay.

## Contract changes

- negotiation priority is now documented in `docs/contracts.md` and pinned by a test
- FourCCs are normalized at ingest, so `facelock devices --json` reports `"Y16"` rather than `"Y16 "`. Direct backend only: the D-Bus `DeviceInfo` carries no formats field, so a daemon-backed `--json` reports `"formats": []` either way
- a device with a padded stride that previously decoded sheared frames now fails at open

## Files

| Path | Why it matters |
|---|---|
| `crates/facelock-camera/src/capture.rs` | format selection, stride check, Y16 scale pinning *(start here)* |
| `crates/facelock-camera/src/device.rs` | decodability filter and its interaction with IR classification |
| `crates/facelock-camera/src/preprocess.rs` | NV12 and Y16 converters, shared YUV math |
| `crates/facelock-camera/src/quirks.rs` | `y16_bit_depth` quirk key |
| `docs/security.md` | what the pinned scale protects, and where it does not |
| `docs/compatibility.md` | IPU6/IPU7 recipe and the IR situation on that hardware |

## Reviewer notes

- **A scene-derived Y16 scale has a fail-open direction, and it ships with this branch.** `ir_texture_min_stddev` is an absolute cutoff on a scale-dependent statistic. A shift below true acts as a contrast stretch on exactly the buffer `check_ir_texture` measures, and can carry a flat spoof past the cutoff. Documented in `docs/security.md` with its preconditions; the generic fix is #161. Reachable only on Y16 devices with no declared `y16_bit_depth`, which today cannot decode at all — Y16 is not in main's format list, so RealSense IR is currently non-functional rather than insecure.
- **V4L2 cannot report a sensor's effective bit depth**, so the shift on an unlisted camera is necessarily inferred. `v4l2_pix_format` carries `bytesperline`, which describes the container, not the valid bits. This is why the quirk key exists and why quirks alone do not generalize.
- **The calibration burst treats any dequeue error as fatal.** `v4l`'s capture stream advances `arena_index` only on a successful dequeue and re-queues that index on the next call, so one failure wedges the stream with EINVAL. Tolerating errors would have returned `Ok` for a camera whose every capture fails.
- **Rebased over 159 commits.** The four original commits replay; `a7dd419` carries the part that could not be expressed as a conflict resolution — reconciling the decodability filter with the evidence-based IR classification from #98/#99, where the two format lists deliberately disagree on Y8/Y10/Y12.

## Validation

- `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt`
- container tiers: PAM smoke, one-shot end-to-end, state layout, daemon integration
- one-shot tier covers both directions of the `require_ir` gate on a multi-node camera, and that the refusal is the IR gate rather than an earlier pre-flight rejection
- Logitech BRIO: auto-detection selects the IR node, negotiates GREY, IR emitter toggles, live capture and the `--ignored` hardware suite
- NV12 end-to-end on IPU7 hardware is unverified — no such device available

Part of #89

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -280,6 +280,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   negotiating an unsupported format (and then failing every capture), opening a
   device with no decodable format errors immediately with the advertised and
   supported format lists.
+- **Pixel-format names lose V4L2's trailing-space padding on machine-readable
+  surfaces** (#89): FourCCs are normalized where they enter facelock (device
+  enumeration and quirks-file parsing), so `facelock devices --json` and the
+  `ListDevices` D-Bus reply now carry `"Y16"` where they previously carried
+  `"Y16 "`. The human-readable `facelock devices` table already trimmed and is
+  unchanged. A consumer matching the padded spelling exactly needs updating; one
+  that trims already does not.
 
 ### Fixed
 
@@ -404,6 +411,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (hardware truth, no frame inspected); otherwise it is calibrated from the peak
   of a short burst of frames rather than a single frame, so a dark pre-AGC frame
   at open no longer pins an 8-bit scale that clips the rest of the session.
+  "The session" is the lifetime of one open camera: the daemon's warm camera
+  hold keeps the scale it pinned, and a reopen recalibrates — no scale is
+  carried across a reopen onto a device that did not produce it. On IR hardware
+  the burst is emitter-LED-on time inside `Camera::open` (bounded by one
+  second); declaring `y16_bit_depth` skips it.
 
 ## [0.1.4] - 2026-05-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -280,13 +280,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   negotiating an unsupported format (and then failing every capture), opening a
   device with no decodable format errors immediately with the advertised and
   supported format lists.
-- **Pixel-format names lose V4L2's trailing-space padding on machine-readable
-  surfaces** (#89): FourCCs are normalized where they enter facelock (device
-  enumeration and quirks-file parsing), so `facelock devices --json` and the
-  `ListDevices` D-Bus reply now carry `"Y16"` where they previously carried
-  `"Y16 "`. The human-readable `facelock devices` table already trimmed and is
-  unchanged. A consumer matching the padded spelling exactly needs updating; one
-  that trims already does not.
+- **Pixel-format names lose V4L2's trailing-space padding** (#89): FourCCs are
+  normalized where they enter facelock (device enumeration and quirks-file
+  parsing), so `facelock devices --json` now carries `"Y16"` where it
+  previously carried `"Y16 "`. This applies on the direct backend only, which
+  is the one that reports formats at all: the D-Bus `DeviceInfo` carries no
+  formats field, so a daemon-backed `--json` reports `"formats": []` either
+  way. The human-readable `facelock devices` table already trimmed and is
+  unchanged. A consumer that reads formats from the direct backend and matches
+  the padded spelling exactly needs updating; one that trims already does not.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,8 +120,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Intel IPU6/IPU7 processed cameras via v4l2-relayd) and Y16 (16-bit IR grayscale,
   bit-depth-aware conversion) are decoded natively. Negotiation priority is now
   `GREY > Y16 > YUYV > NV12 > MJPG`. IPU6/IPU7 relay cameras can now be opened,
-  previewed and enrolled — but they expose no Linux-reachable IR sensor, so
-  authentication still fails under the default `security.require_ir = true`.
+  previewed and enrolled — but their IR sensors have no in-tree Linux driver
+  yet, so authentication still fails under the default
+  `security.require_ir = true` on stock kernels (experimental community driver
+  support is tracked in #101).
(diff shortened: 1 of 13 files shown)


tyvsmith:
Review fixes applied in c653af2 (principal review, all findings verified before fixing):
- **Y16 scale pinned at camera open** (security): shift derived once from a calibration frame, stored on `Camera`, never recomputed — per-frame max derivation was attacker-modulatable contrast normalization upstream of the IR texture check, and a single specular glint blacked out whole frames. Noted in `docs/security.md`.
- NV12 odd-width OOB panic fixed (UV stride = `2*ceil(w/2)`), with 3x3 regression test.
- Padded strides (`bytesperline` > row size) rejected at open with both values named, instead of decoding sheared frames.
- FourCC normalized once at the query boundary; `DECODABLE_FORMATS` order pinned by test (changing it now requires a conscious contracts.md update).
- Quirk `format_preference` naming an undecodable format is warned and ignored instead of winning negotiation.
- `nv12_to_rgb` returns `Option`; excluded IR nodes are warn-logged with their formats; CHANGELOG no longer oversells (IPU relay ≠ IR); `cargo fmt` fixed.

Resolved judgment call: kept the hand-rolled YUV converters per the repo's dependency-minimalism rule (cross-converter equivalence test pins the math).

tyvsmith:
### New info on the `SGRBG10` exclusion in this PR (no change requested — merge as is)

Read the fix pack's [`howdy/ir_reader.py`](https://github.com/jibsta210/svp7500-camera-fix-pack/blob/main/howdy/ir_reader.py) and [`IR-FINDINGS.md`](https://github.com/jibsta210/svp7500-camera-fix-pack/blob/main/IR-FINDINGS.md). They pin down the format this PR's auto-detect filter deliberately skips, and it turns out to be more interesting than "a raw Bayer sensor node".

**The `SGRBG10` node on IPU7 is the IR node.** The HM1092 is physically **monochrome**; the driver tags it `SGRBG10` (`'BA10'`) only because that is what the Windows IPU graph settings declare. Measured layout: 648x368, **`bytesperline = 1344`** (672 px of stride for 648 active), 16-bit LE containers holding 10-bit values, decoded as `crop to width` then `>> 2`. Honouring the Bayer tag is what breaks it — libcamera's SoftwareIsp debayers mono data, and the PipeWire path additionally renegotiates to 1920x1080 and lands the payload in a mostly-black buffer.

So the two behaviours this PR adds both correctly exclude the node **today** (we genuinely cannot decode it yet), and both will need a carve-out later:

1. `DECODABLE_FORMATS` has no `BA10`, so auto-detection skips it — the PR description's "raw Bayer sensor nodes (e.g. IPU7's `SGRBG10`) can no longer be auto-selected" is accurate but reads as if the node is uninteresting. It is the IR sensor.
2. `expected_stride()` fail-fasts on the padded stride: 648 x 2 = 1296 != 1344.

The decode itself is essentially already here: 10-bit-in-16-bit-LE `>> 2` is this PR's Y16 path with `y16_bit_depth = 10`. What is missing is a stride-aware arm plus three subsystems facelock does not have at all (no references to `media-ctl`, `v4l-subdev`, or `/sys/class/leds` anywhere in the workspace; `ir_emitter.rs` is UVC-XU-only):

- [ ] `BA10`/SGRBG10-as-mono decode arm, routed through the existing Y16 conversion with the shift pinned by `y16_bit_depth = 10`
- [ ] Stride-aware capture — support padded `bytesperline` (crop per row) instead of rejecting it, for formats where padding is expected
- [ ] **Media-graph configuration** (`VIDIOC_SUBDEV_S_FMT` + `MEDIA_IOC_SETUP_LINK`, or `media-ctl`, or a boot-time oneshot). On a fresh boot the CSI2 -> capture link is **disabled** and the CSI-2 pads sit at 4096x3072, so `VIDIOC_STREAMON` yields zero frames. The `v4l` crate does not cover subdevs. New subsystem, not a format arm.
- [ ] **Entity-name device resolution** — `/dev/videoN` and `/dev/v4l-subdevN` shuffle between boots on IPU7, and the CSI-2 port differs per machine (port 2 on XPS 16 DA16260, port 1 on Dell Pro 14)
- [ ] **sysfs LED-class emitter backend** for `/sys/class/leds/HIMX1092_00::ir_flood_led/brightness`, with guaranteed off-on-`Drop` (their Python version leaked the illuminator **on** after every auth when `os.open` was torn down during interpreter shutdown). INT3472 owns the GPIO, so the sensor driver reports `ir_led=none` and cannot drive it.
- [ ] `force_ir` quirk on a `hm1092`/`HIMX1092` name pattern
- [ ] No rotation quirk — the sensor applies H+V flip **in hardware** (reg `0x0101=0x03`); only the RGB relay needs rotate-180

Full detail, including the `link_frequency` = 180,480,000 prerequisite and what still needs hardware confirmation, is in #101.

Nothing here blocks this PR — the exclusion is the right behaviour until the decode arm exists. Filing it so the next person who reads `DECODABLE_FORMATS` does not conclude the IPU7 IR node is a dead end.

PR #91, opened 2026-08-08
fix(test): harden container pamtester download, host-network builds, model preflight

tyvsmith:
## Summary

Container-test infrastructure fixes found while validating the #89 series locally:

- **pamtester download hardened**: `curl -sL | tar` piped truncated/error responses straight into gzip, failing the image build with an opaque `gzip: stdin: unexpected end of file`. Now `curl -f` + retries + a sha256 check.
- **`podman build --network=host`** in `_build-test-container`: rootless podman's default build network (pasta/slirp) truncated the SourceForge redirect chain deterministically on at least one host; build-time isolation buys nothing here.
- **`_require-models` preflight for the camera tiers**: the Containerfile bakes `models/` into the image with a tolerant `|| true`, so building from a fresh clone/worktree (the ONNX models are downloaded, not tracked) silently produced an image whose daemon cannot load the face engine and whose enroll bails before opening the camera. `test-arch-integration`/`test-arch-oneshot` now fail up front with a clear message instead.

## Validation

- Stock `just test-arch-pam` previously failed at the pamtester step on this host; with these changes it builds and passes 22/22.
- `just test-arch-oneshot` / `just test-arch-integration` run to completion with models present (verified with a human in frame).
- The preflight was exercised for real: a fresh worktree without models now errors immediately instead of producing the misleading in-container failures.

Independent of the #89 code PRs, but the local validation of that series depends on it.

### DIFF
--- a/justfile
+++ b/justfile
@@ -50,8 +50,23 @@ _build-test-container: build-release
 test-arch-pam: _build-test-container
     podman run --rm facelock-pam-test
 
+# Guard for the camera tiers: the Containerfile bakes models/ into the image
+# with a tolerant `|| true`, so building from a checkout without the ONNX
+# models (fresh clone/worktree — they are downloaded, not tracked) produces an
+# image whose daemon cannot load the face engine and whose enroll bails before
+# opening the camera. Fail loudly here instead of debugging opaque test FAILs.
+_require-models:
+    #!/usr/bin/env bash
+    set -euo pipefail
+    if ! ls models/*.onnx >/dev/null 2>&1; then
+        echo "error: no ONNX models in models/ — the camera test tiers need them baked into the image." >&2
+        echo "       Copy them from an existing checkout or download via 'sudo facelock setup'," >&2
+        echo "       e.g.: cp /path/to/other/checkout/models/*.onnx models/" >&2
+        exit 1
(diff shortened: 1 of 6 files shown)


tyvsmith:
Review fixes applied in a3a61d7:
- **pamtester install deduped**: one hardened `test/install-pamtester.sh` (curl -f + retries + sha256 pin) now used by all four Containerfiles — the deb-e2e, deb-tpm-e2e, and rpm-e2e files previously kept the unhardened download this PR set out to eliminate.
- `_require-models` checks the two required models by name (`scrfd_2.5g_bnkps.onnx`, `w600k_r50.onnx`) instead of a glob that optional-only checkouts satisfied; the remedy text now points at `cp /var/lib/facelock/models/*.onnx models/` (where setup actually downloads to).
- The `--network=host` comment states the real constraint: **empirically, the failure is a TLS handshake error (curl exit 35) under podman's default rootless network, reproducible and independent of the URL used** — so the reviewer's direct-mirror alternative was tested and does not remove the need for host networking. The flag stays; the checksum still guarantees integrity.
- Reciprocal keep-in-sync comment added for the ci.yml duplication (recipe restructuring deferred).

PR #92, opened 2026-08-08
fix(enroll): D-Bus timeout above server deadline + failure breakdown + direct-mode unification

tyvsmith:
## Summary

Second PR for #89 — enrollment robustness.

- **D-Bus Enroll timeout race fixed**: the CLI's fixed 15s method timeout was at or below the daemon's enrollment deadline (`3 x recognition.timeout_secs`, min 15s), so daemon-mode enrollment could die client-side with `I/O error: timed out` while the daemon was still enrolling. `Config::enroll_timeout_secs()` is now the single source of truth; the new `send_enroll()` uses a dedicated connection with that deadline + 15s margin. The shared 15s proxy still serves every other method; the PAM auth path is untouched. Contract documented in `docs/contracts.md`.
- **Enrollment failure breakdown**: when too few frames are captured, the error reports why (`— rejected frames: 42 too dark, 3 no face ... (last: <capture error>)`) plus a remediation hint when one cause dominates. Resolves the issue's "insufficient face captures with no explanation" finding.
- **Direct-mode enrollment unified with the daemon loop** (found during hardware validation): `direct.rs` carried a drifted local copy of the enrollment loop that skipped the frame-quality gate and angle-diversity check and would have hidden the new breakdown. Both modes now share `facelock_daemon::enroll`.
- **Direct mode no longer D-Bus-activates the daemon**: label auto-generation and the model-count warning made unconditional D-Bus `ListModels` calls, which could boot the system daemon via bus activation and silently flip enrollment from direct to daemon mode. They now read the store directly when direct mode applies.

## Validation

- Clippy `-D warnings` clean; full workspace tests pass (new: `enroll_timeout_secs` formula, rejection-summary/hint tests). Container PAM tier 22/22; oneshot + integration camera tiers pass with a human in frame.
- Hardware: covered-camera enroll in direct mode yields the breakdown + lighting/no-face hint; with `timeout_secs = 10` and a live daemon, the client now waits the full ~30s deadline and receives the daemon's breakdown over D-Bus instead of timing out at 15s (reproduced the exact failure from the issue, then confirmed fixed); an uncovered enroll through the unified loop stores 10 embeddings normally.

Note: `direct.rs` still contains a `facelock_daemon_auth` twin of the drift hazard on the auth path — deliberately untouched here (security-critical path, needs its own review); follow-up PR incoming.

Merge note: touches `crates/facelock-cli/src/ipc_client.rs` in a region adjacent to #93's changes — merge this before #93 for a clean rebase. CHANGELOG conflicts across the series are trivial (combine Unreleased bullets).

Part of #89


tyvsmith:
Review fixes applied in e42198b:
- **Failed enrollment no longer leaves a partial template live in the DB** — both the angle-diversity and insufficient-captures failure paths now delete the model row they created; regression test proves the fix by failing when it's removed. (Pre-existing bug; the unification made it reachable in both modes, so it was fixed here.)
- Dead config-blind `send_enroll(…, 15)` fallback arm deleted (now bails pointing at `send_enroll`); `send_enroll` takes `&Config` and derives the deadline itself, making the misuse inexpressible.
- Face-engine errors split from camera errors (own counter, own hint) so ONNX failures no longer point users at `device.path`.
- contracts.md formula now reads `3 × max(recognition.timeout_secs, 5)`; majority-boundary hint tests added; unreachable match arm removed; `cargo fmt` fixed.

Deferred (reviewed, intentionally not here): startup-margin scaling beyond +15s, surfacing `quality_hint` per-frame tallies, moving hint prose from daemon errors to the presentation layer, caching the mode decision per command.

PR #93, opened 2026-08-08
feat(setup): manage facelock group membership + actionable AccessDenied hint

tyvsmith:
## Summary

Third PR for #89 — permissions UX.

The D-Bus system-bus policy admits only root and the `facelock` group, but nothing ever added the human user to the group: packaging creates the group, `sudo facelock setup` ran entirely as root and only errored if the group was missing, and dev installs masked the gap (`just install` runs `usermod -aG`). A normal user's first `facelock preview`/`test` after setup failed with a bare `AccessDenied`.

- **Setup manages membership**: creates the `facelock` system group if missing (source installs) and adds the invoking user (`SUDO_USER`/`DOAS_USER`) — the wizard asks first (default yes), non-interactive mode adds unconditionally to match `just install`, and both print a log-out/log-in reminder. Runs before `secure_setup_paths`, which needs the group to exist.
- **Actionable AccessDenied**: the CLI detects D-Bus AccessDenied anywhere in the error chain (bus-policy `MethodError` or FDO variant) and appends the group-membership fix. The bus-level deny fires before the daemon's own friendly check can, so this is the only place the hint can live.
- Troubleshooting docs updated with the `facelock preview` AccessDenied symptom.

## Validation

- Clippy `-D warnings` clean; full workspace tests pass (new: AccessDenied hint mapping tests). Container PAM tier 22/22.
- Container: fresh Arch container with neither group nor member — `SUDO_USER=tester facelock setup --non-interactive` creates the group, adds the user, prints the reminder; `id tester` confirms.
- Host: `sudo -u nobody facelock list` against a running daemon shows the hint instead of bare AccessDenied; the wizard correctly reports an already-enrolled member.

Merge note: merge after #92 (adjacent additive changes in `ipc_client.rs`; trivial rebase). CHANGELOG conflicts across the series are trivial (combine Unreleased bullets).

Part of #89


tyvsmith:
Review fixes applied in 5d5bf91:
- **The AccessDenied hint now reaches its motivating commands**: `preview --text-only` prints the error instead of silently exiting, the Wayland preview surfaces it once instead of looping on stale frames, and `facelock test` distinguishes query-failure from "no models enrolled" (which previously misdiagnosed a denial as a missing enrollment).
- Hint wording softened to compose with daemon-originated denials ("some operations require root regardless of group membership") so it no longer contradicts precise root-only messages.
- Name matching extracted to a pure `is_access_denied_name` helper with direct tests; the FDO-variant test no longer claims to cover the wire path.
- Doc comment/CHANGELOG no longer claim unconditional adds (no-op without SUDO_USER/DOAS_USER is documented); contracts.md setup entry mentions group management; `facelock devices` error ordering keeps the hint as the headline; `cargo fmt` fixed.

Deferred (reviewed, tracked for follow-up): `--no-group`/`--group-user` escape hatch for the non-interactive add; busybox `addgroup`/`adduser` fallback; justfile/sysusers group-creation dedup.

PR #94, opened 2026-08-08
fix(direct): unify direct-mode auth with the daemon implementation

tyvsmith:
## Summary

`crates/facelock-cli/src/direct.rs` carried a local fork of the authentication loop (`mod facelock_daemon_auth`) that had drifted from the real `facelock_daemon::auth::authenticate_with_embeddings`. Direct mode now calls the daemon crate's function directly, so both modes share one implementation.

This follows the same pattern as the enroll unification in the #89 PR series (commit 75ad2d8): delete the duplicated module, call the daemon crate's public function. The signatures were already identical, so this is a drop-in call-site swap — **no change to `facelock-daemon`'s public API**, and no documented contract in `docs/contracts.md` is affected.

## Scope: this is not the PAM path

The drifted copy was reachable only from **`facelock test` in oneshot/direct mode** (`commands/test_cmd.rs`). The PAM authentication path is unaffected: `facelock auth` (`commands/auth.rs`) already called `auth::authenticate_with_embeddings` and does its own `pre_check`-equivalent inline. So the drift degraded diagnostics and observability, not the security of real authentications.

## Behavioral diff found

**No security check was missing from the fork.** Every P0/P1 gate from `docs/security.md` was present in both copies — IR texture on the raw frame (§1.C, correctly on `frame.gray`, not CLAHE), the frame-variance sliding window (§1.B, including `variance_ever_passed` and the `VarianceNotSatisfied` reason), landmark liveness, dark-frame handling, device coupling (§1.D), and constant-time matching via `best_match`. The drift was in the audit/observability and memory-hygiene layers:

| # | Difference | Impact |
|---|---|---|
| 1 | **No audit entries written.** The daemon writes `success` / `failure` / `error: all frames dark` entries to the configured audit log; the fork wrote none. | Direct-mode attempts left no audit trail (`docs/security.md` §5.A audit-trail completeness). Default `audit.enabled = false`, so this is latent until an operator turns it on. |
| 2 | **Device-filtered-out templates never zeroized.** The daemon zeroizes the full embedding set after building the device-coupled compare set; the fork took an immutable slice, cloned only the allowed entries, and zeroized just those. | Templates skipped by device coupling stayed resident in memory (`docs/security.md` P2, memory zeroing). |
| 3 | **`[snapshots]` ignored entirely.** No `save_snapshot` call on either the success or failure path. | The config key silently did nothing in direct mode. Default `mode = "off"`. |
| 4 | **Legacy-template nudge missing.** The daemon's `info!` for pre-V6 / NULL `device_id` templates ("re-enroll to couple them to this camera") was absent. | Users on legacy templates got no prompt to re-enroll. |
| 5 | **Thinner per-frame diagnostics.** Missing `"dark frame, skipping"`, `"no faces detected"`, and the per-face `"face comparison"` similarity line; the success log omitted the `matched` frame count, and the no-variance branch lacked the `(no variance check)` marker. | `facelock test` is *the* tuning command, so these are exactly where the debug output matters most. |

Both newly-inherited behaviors (audit, snapshots) default to off and are best-effort — `write_audit_entry` and `save_snapshot` log failures and never propagate, so neither can fail an authentication.

## Deliberately preserved call-site difference

Direct-mode `facelock test` still does **not** run `auth::pre_check`, and that is intentional, not an oversight. `pre_check` enforces rate limiting and the SSH/lid aborts; wiring it into the diagnostic would mean five `facelock test` runs exhaust a user's real face-auth budget, and that debugging over SSH becomes impossible. Note that daemon-mode `facelock test` *does* go through `pre_check` (via `DaemonRequest::Authenticate`), so the two modes still differ here — including on `require_ir`, which direct-mode `facelock test` does not enforce. That is a pre-existing divergence in the diagnostic command's semantics and a product decision rather than a mechanical unification; flagging it for a separate call rather than changing it silently here. The container suite runs `facelock test` repeatedly, which is evidence that repeated diagnostic runs are the expected usage.

`security.require_ir` remains `true` by default and no security default was weakened.

## Validation

- `cargo build --workspace` — clean
- `cargo test --workspace` — **453 passed, 0 failed** (includes the new `failed_auth_writes_audit_entry` regression test)
- `cargo clippy --workspace -- -D warnings` — clean
- Container PAM tier (`test/Containerfile`) — **22 passed, 0 failed**

Deleting the fork also means direct mode now inherits the daemon's existing regression coverage in `crates/facelock-daemon/tests/daemon_integration.rs`: `device_mismatch_never_reaches_success`, `legacy_null_device_id_still_authenticates`, `static_matching_frames_report_variance_reason`, and `still_then_moving_frames_recover_and_authenticate`.

Net diff: 69 insertions, 295 deletions.

Related to #89


tyvsmith:
Review fixes applied in ae3d607:
- **Audit entries now carry their source**: `AuditSource` enum (`daemon`/`oneshot`/`test`, closed set so new call sites must classify themselves) stamped on every entry `authenticate_with_embeddings` writes plus the daemon pre_check rejections and the oneshot storage-error path. Optional field with `skip_serializing_if`, so pre-upgrade log lines remain valid; documented in contracts.md. This closes the forensic ambiguity where a `facelock test` success (which skips pre_check gates) was indistinguishable from a policy-approved authentication.
- Regression-test comment now states precisely what it pins (the daemon function's audit contract) and that a re-fork of direct.rs would not be caught by it.
- `direct.rs` comment keeps "do not re-fork", drops the drift-prone behavior enumeration; Copilot's path correction applied; `cargo fmt` fixed.
- The extended test asserts both a `daemon` and a `test` line against one log, and was mutation-checked (`source: None` fails it).

Deferred to filed issues: the pre_check mode-dependence (#96) and the `commands/auth.rs` pre_check fork (#95 — the drifted copy on the PAM path).

PR #112, opened 2026-08-14
feat(cli): root-by-default privilege map; root-only test with safe semantics (DEC-6, N11)

### LINKED ISSUE #96: facelock test enforces pre_check gates only when the daemon happens to be up
Found during the #89 review series (principal review of #94).

## Problem

`should_use_direct` (`crates/facelock-cli/src/ipc_client.rs:51`) returns true when `daemon.mode = oneshot` **or when the daemon simply isn't answering on D-Bus**. Daemon-mode `facelock test` routes through `DaemonRequest::Authenticate` → `pre_check` (require_ir, rate limiting, suppress_unknown); direct-mode `facelock test` calls `authenticate_with_embeddings` with **none** of those gates (`require_ir` and `RateLimiter` appear nowhere in the direct/test path).

So the same command on the same config enforces the security gates or skips them depending on whether the daemon process happens to be running. An operator can validate enrollment with `facelock test`, see it pass, and conclude `require_ir` is satisfied — while the run bypassed it entirely.

After #94 unified the core loop, this exception is invisible: `direct.rs` now calls the canonical function, which sets the expectation that direct mode *is* the daemon path.

## Suggested fix

Decide explicitly: if the diagnostic must not consume rate-limit budget (defensible — the container suite runs it repeatedly, and debugging over SSH must stay possible), make that an explicit parameter or a documented `pre_check` variant, and say so in `docs/security.md`. `facelock test` output could also state which gates were skipped.

Note: #94 added a `source` field to audit entries which mitigates the forensic side of this.

tyvsmith:
## Summary

Implements DEC-6 (CLI is root-by-default) and N11 (`facelock test` is root-only with safe rate-limit semantics), closes #96.

## Privilege table (as shipped)

**Unprivileged** (unchanged, pinned by tests): `is-enrolled`, `hyprlock …`, `config` (display), `--help`/`--version`.

**Root — interactive prompt, hard error when non-interactive/scripted** (`ipc_client::require_root`): `setup`, `enroll`, `test`, `preview`, `bench`, `encrypt`, `decrypt`, `tpm`, `reseal`, `restart`, `config --edit`, `remove`, `clear`, `list`, `status`, `devices`.

**Root — hard error only, never a prompt** (`ipc_client::require_root_scripted`): `audit` (`daemon`'s own equivalent check predates this PR and was not touched).

## Per-command delta

| Command | Before | After |
|---|---|---|
| `status` | No root check; D-Bus calls (`Ping`) already root-only on the wire → bare AccessDenied | `require_root` as the first statement, before any output |
| `devices` | No root check; `ListDevices` already root-only on the wire | `require_root` as the first statement |
| `preview` | No root check; `PreviewDetectFrame` already root-only on the wire | `require_root` as the first statement (privilege gate only — preview/text_only/render/wayland_preview submodules untouched) |
| `bench` | Root required only for `{cold-auth,warm-auth,calibrate,report}` **and** only when `encryption.method == tpm` | Unconditional `require_root` at the top of `run()`, all subcommands, superseding the old conditional |
| `config --edit` | No root check; display and edit shared the same unprivileged path | `require_root` before `open_in_editor`; display (`config`, no flag) stays unprivileged |
| `audit` | No root check | `require_root_scripted` (hard error only, never prompts — audit is typically scripted) |
| `list` | Direct-mode fallback only required root when the DB was unreadable; D-Bus branch had no proactive check (bare AccessDenied) | Proactive `require_root` at the top of `run()` for both transports; simplified `fetch_models`'s now-dead retry-after-AccessDenied branch |
| `remove` | **C6 bug**: Y/N confirmation ran before the root check (confirmed root-only historically) | Root check moved to the first statement, before the confirmation prompt |
| `clear` | Already correct (root check first) | Verified only — no change |
| `test` | Root required only on the direct branch (`ipc_client::require_root` at old line 87), after the models-exists prompt; D-Bus branch had no check at all | Root check is the first statement in `run()`, before the models-exists check/prompt; see N11 below for transport-specific semantics |

## N11 — `facelock test` design decisions

- **Root-only, both transports.** `require_root("sudo facelock test")` is the first statement in `test_cmd::run`.
- **Full detail preserved.** Root gets real similarity scores on the wire (`AuthResult.similarity` is redacted only for non-root D-Bus callers, from Wave-1 PR-C); nothing is stripped for `test`.
- **`pre_check` now applies on the direct branch** (previously it was explicitly skipped — see the removed doc comment on `direct::authenticate`). Added `PreCheckContext` (`crates/facelock-daemon/src/auth.rs`) with `skip_ssh_gate`/`skip_lid_gate`, and `pre_check_with_context`/`pre_check_audited_with_context`. `pre_check`/`pre_check_audited` keep their old signatures as thin wrappers over `PreCheckContext::enforced()`, so this is purely additive — zero signature change for existing callers (`commands/auth.rs`'s oneshot path, `handler.rs`'s daemon path — neither was touched).
  - The override is scoped to the **direct** transport only: `direct::authenticate` resolves the live camera device (cheap, no camera I/O) and calls `pre_check_audited_with_context(.., PreCheckContext::test())` before opening the camera.
  - It's intentionally **not** duplicated on the daemon transport: `test` there is just another `Authenticate` D-Bus call, indistinguishable from real auth at that layer. `abort_if_ssh` checks the *daemon process's* environment (never an SSH session) so it's already inert there; `abort_if_lid_closed` reads a system-wide hardware file, so it still applies uniformly — documented as an intentional asymmetry in `docs/contracts.md`, not a bug.
- **Rate-limit consumption exempted; the check is not.**
  - Direct: `direct::authenticate` structurally never calls `RateLimiter::record_failure` (it never did — this is preserved, not new).
  - Daemon: **the one authorized `facelock-daemon` touch.** Refactored `Handler::handle()`'s `Authenticate` arm into `pub fn handle_authenticate(&mut self, user: String, charge_failed_attempt: bool)`; `handle()` calls it with `true` (zero behavior change for every existing caller, including `facelock-daemon/tests/daemon_integration.rs`'s two `handle(DaemonRequest::Authenticate{..})` calls). `commands/daemon.rs`'s `authenticate()` D-Bus method now calls `handler.handle_authenticate(user.clone(), !caller_is_root)` instead of `handler.handle(DaemonRequest::Authenticate{..})` — `caller_is_root` was already computed there (for similarity redaction, Wave-1 PR-C), so this is a one-line swap at the single call site, not new plumbing.
  - The rate-limit **check** (is `user` already over budget?) is unaffected on both transports — an already-limited user's `test` run reports "rate limited" exactly like real auth. Root can already clear the limiter directly (it owns the database), so exempting *consumption* costs nothing security-wise while still surfacing an existing lockout instead of masking it.

### The declared facelock-daemon touch (and its one necessary companion edit)

- `crates/facelock-daemon/src/auth.rs`: additive only (`PreCheckContext` + two new `_with_context` functions; old functions become wrappers).
- `crates/facelock-daemon/src/handler.rs`: `Authenticate` arm extracted into `pub fn handle_authenticate(user, charge_failed_attempt)`; `handle()`'s dispatch becomes a one-line delegation. No other arm touched.
- `crates/facelock-cli/src/commands/daemon.rs`: **one line** in the `authenticate()` D-Bus method (not in my explicit ownership list, called out here per the "narrowly-authorized extra" instruction) — swaps `handler.handle(DaemonRequest::Authenticate{..})` for `handler.handle_authenticate(user.clone(), !caller_is_root)`. This is structurally unavoidable: `caller_is_root` is only knowable at the D-Bus layer (from `GetConnectionUnixUser`), and the daemon library has no visibility into caller identity. Blast radius: exactly one call site, one of `commands/daemon.rs`'s ten `handler.handle(...)` calls (the other nine — Enroll, ListModels, RemoveModel, ClearModels, PreviewFrame, PreviewDetectFrame, ListDevices, ReleaseCamera, Ping, Shutdown — untouched); no change to authorization, redaction, or any other method's behavior.
- `crates/facelock-cli/src/direct.rs`: not in my explicit ownership list either, but `direct::authenticate` is the *only* function backing `test`'s direct-mode path, and N11 can't be wired correctly without editing it. Rewrote it to resolve the camera device once for `pre_check`'s `device_is_ir`, call `pre_check_audited_with_context`, and removed the now-redundant manual `has_models` early return (pre_check's own not-enrolled/`suppress_unknown` branch supersedes it) — net simplification, not just an addition. No other function in the file was touched.

## Tests

- `crates/facelock-daemon/src/auth.rs`: `PreCheckContext::test()`/`enforced()` field pins; `pre_check_with_context` skips the SSH gate under `test()` and still blocks it under `enforced()`; `pre_check()` (no context) stays equivalent to `enforced()` (regression pin so real auth paths can't silently pick up a relaxed gate). SSH-env-mutating tests share a `Mutex` to avoid flaking against parallel test threads (4 tests in this file now touch `SSH_CONNECTION`, up from 1).
- `crates/facelock-daemon/tests/daemon_integration.rs`: `exempted_authenticate_does_not_consume_rate_limit_budget` — three failed exempted attempts against `max_attempts = 1`, then a direct `FaceStore::check_rate_limit` assertion that the table is untouched.
- `crates/facelock-cli/src/ipc_client.rs`: `require_root_scripted` never-prompts structural test + root-pass test; `add_access_denied_hint` root-required vs. group-membership hint tests.
- `crates/facelock-cli/tests/cli_smoke.rs`: ten new black-box subprocess tests (`status`, `devices`, `preview`, `test`, `audit`, `bench`, `config --edit`, `list`, `remove`, `clear`) — each spawns the real `facelock` binary non-root with stdin closed and asserts it refuses with "Root required" *before* emitting any command-specific output (C6 contract). Skips gracefully (not a failure) if the test runner happens to be root.

## Gates

```
just check          → PASS (cargo test/clippy/fmt-check/audit all green; 216 facelock-cli unit tests, 15 cli_smoke integration tests, 57 facelock-daemon unit tests, 20 daemon_integration tests, all passing)
just test-arch-pam        → 22/22 PASS
just test-arch-layout     → 21/21 PASS (is-enrolled's unprivileged group-member behavior unaffected)
```

## Cross-PR notes

- `list.rs`'s D-Bus branch previously relied on a bare `AccessDenied` (fixed generically by the hint change); it now proactively prompts like every other command, for consistency — not strictly required by DEC-6's letter but a clear UX improvement with no correctness downside.
- `status.rs`/`test_cmd.rs` have pre-existing messaging bugs (documented in the Aug-2026 architecture review, e.g. `test_cmd.rs`'s "No face models enrolled" wording when it can't read the store) that are out of scope here — flagged for the next PR.
- `Cargo.lock` picked up a handful of dropped `async-io`/`async-process`-family entries that were no longer reachable by the dependency graph (pure pruning, no version bumps of anything in use) — a byproduct of the first build in this worktree, not an intentional dependency change.

## Docs to reconcile (not touched here — book/ and man/ are out of my ownership for this PR)

- `man/facelock.1`: `test`, `status`, `devices`, `preview`, `bench`, `config --edit`, `audit` entries should note the new root requirement and (for `audit`) the no-prompt behavior.
- `book/src/cli-reference.md`: same set of commands.
- `book/src/troubleshooting.md`, `docs/troubleshooting.md`, `book/src/security.md`, `docs/security.md`: all still say "join the facelock group" for AccessDenied-style errors in places that are now root-only denials; worth a pass once this lands.

---
base is integration/wave-1; retarget to main after #107-#111 merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,67 +192,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-executor"
-version = "1.14.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c96bf972d85afc50bf5ab8fe2d54d1586b4e0b46c97c50a0c9e71e2f7bcd812a"
-dependencies = [
- "async-task",
- "concurrent-queue",
- "fastrand",
- "futures-lite",
- "pin-project-lite",
- "slab",
-]
-
-[[package]]
-name = "async-io"
-version = "2.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
-dependencies = [
- "autocfg",
- "cfg-if",
- "concurrent-queue",
- "futures-io",
- "futures-lite",
- "parking",
- "polling",
- "rustix 1.1.4",
- "slab",
- "windows-sys 0.61.2",
-]
-
-[[package]]
-name = "async-lock"
-version = "3.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
-dependencies = [
- "event-listener",
- "event-listener-strategy",
- "pin-project-lite",
-]
-
-[[package]]
-name = "async-process"
-version = "2.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc50921ec0055cdd8a16de48773bfeec5c972598674347252c0399676be7da75"
-dependencies = [
- "async-channel",
- "async-io",
- "async-lock",
- "async-signal",
- "async-task",
- "blocking",
- "cfg-if",
- "event-listener",
- "futures-lite",
- "rustix 1.1.4",
-]
-
-[[package]]
 name = "async-recursion"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -261,24 +200,6 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
-]
-
-[[package]]
-name = "async-signal"
-version = "0.2.13"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43c070bbf59cd3570b6b2dd54cd772527c7c3620fce8be898406dd3ed6adc64c"
-dependencies = [
- "async-io",
- "async-lock",
- "atomic-waker",
- "cfg-if",
- "futures-core",
- "futures-io",
- "rustix 1.1.4",
- "signal-hook-registry",
- "slab",
- "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4668,12 +4589,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "eee682d202a77e4a9f3b2c2bdf48a7b28af5c08c34ddf66f98c93e5e39464285"
 dependencies = [
  "async-broadcast",
- "async-executor",
- "async-io",
- "async-lock",
- "async-process",
  "async-recursion",
- "async-task",
  "async-trait",
  "blocking",
  "enumflags2",
--- a/crates/facelock-cli/src/commands/audit.rs
+++ b/crates/facelock-cli/src/commands/audit.rs
@@ -5,6 +5,10 @@ use anyhow::{Context, Result};
 use facelock_core::config::Config;
 
 pub fn run(tail: bool, lines: usize) -> Result<()> {
+    // DEC-6: `audit` is typically invoked scripted/non-interactively, so it
+    // hard-errors rather than offering the usual sudo re-exec prompt.
+    crate::ipc_client::require_root_scripted("sudo facelock audit")?;
+
     let config = Config::load()?;
 
     if !config.audit.enabled {
--- a/crates/facelock-cli/src/commands/bench.rs
+++ b/crates/facelock-cli/src/commands/bench.rs
@@ -47,21 +47,10 @@ pub enum BenchCommand {
 }
 
 pub fn run(command: BenchCommand) -> Result<()> {
-    // Auth benchmarks need to decrypt embeddings, which may require root for TPM access
-    let needs_embeddings = matches!(
-        command,
-        BenchCommand::ColdAuth
-            | BenchCommand::WarmAuth
-            | BenchCommand::Calibrate
-            | BenchCommand::Report
-    );
-    if needs_embeddings {
-        if let Ok(config) = Config::load() {
-            if config.encryption.method == facelock_core::config::EncryptionMethod::Tpm {
-                crate::ipc_client::require_root("sudo facelock bench <subcommand>")?;
-            }
-        }
-    }
+    // DEC-6: `bench` is root by default (direct-mode access needs the 0600
+    // root:root database regardless of subcommand, and auth benchmarks may
+    // need TPM access besides). Supersedes the old TPM-only conditional check.
+    crate::ipc_client::require_root("sudo facelock bench <subcommand>")?;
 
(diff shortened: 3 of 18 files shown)


tyvsmith:

PR #113, opened 2026-08-14
fix(auth-path): report failures truthfully — no silent lockouts, fail-closed guards (C2-C5, C7)

### LINKED ISSUE #105: bug: four auth-path failures that report success or the wrong cause
Four places where a failure is reported as something other than a failure. Common shape: an error is collapsed into a value that is indistinguishable from a legitimate result.

**1. Daemon auth silently consumes rate-limit budget until lockout.**
`handler.rs` does `list_models(&user).unwrap_or_default()`. The result feeds `device_allowed_model_ids`, which builds the allowed-ID set by iterating it — so an empty list yields an empty allowed set, `auth.rs` filters the compare set to nothing, and the outcome is a guaranteed "no match". The filter is **unconditional**; `policy.enabled` gates only the warning, so this fires even with device binding disabled (the default). `handler.rs` then records a rate-limit failure. The user is told nothing and retries into a lockout. Two frames up, `pre_check` handles the same failure by propagating.

**2. The pre-keygen guard fails open.**
`setup.rs` has two reads that both fail open — `Err(_) => return Ok(())` on the store open, and `has_any_models().unwrap_or(false)` on the query. The guard exists to prevent minting a new encryption key while encrypted templates exist. Blinded, it concludes "nothing to protect" and proceeds, orphaning every enrollment. Fixing only the first read leaves the guard fail-open.

**3. `facelock clear` can delete nothing and exit 0.**
The same question is answered with opposite defaults per transport: the direct branch uses `.unwrap_or(true)` (assume models exist), the D-Bus branch uses `matches!(...)` which swallows `Err` (assume none). On a transient D-Bus failure the command prints "No face models enrolled" and exits 0 having deleted nothing, while the direct branch on the identical failure proceeds to delete.

**4. `facelock test` states a falsehood.**
`.ok().and_then(...).unwrap_or(false)` then prints "No face models enrolled for user X" — including when the store is present but unreadable. `list.rs` already models the correct three-way discrimination (opened / absent / unreadable).

Related but distinct, already filed: #95 (oneshot mirrors `pre_check` and has drifted), #96 (`facelock test` gate enforcement depends on daemon liveness).

tyvsmith:
Closes #105.

Wave 2, PR F: the confirmed auth-path failure-semantics bugs (gaps C2–C5, C7). Every fix changes a *failure response*, never an auth decision, and every fix ships with the test that pins it.

## The C3 lockout, in one paragraph

`handle_authenticate` folded a `list_models` storage error into an empty model list. An empty model list means an empty device-allowed set, so every stored embedding was filtered out before comparison, the auth loop was guaranteed "no match", and that "failure" was charged against the user's rate limit. The user was told nothing true, retried, and locked themselves out — on the daemon path that every user-facing authentication now takes (post DEC-1). The fix returns `DaemonResponse::Error` (matching what `pre_check` already does for the same failure class two frames up), detected before the camera is acquired. The pinning test injects a failure that passes `has_models` and the embedding load but fails `list_models`, then asserts both halves: the response is `Error`, and the on-disk rate-limit counter is **unchanged** (with `max_attempts = 1`, a single wrongly-recorded failure would flip the follow-up `check_rate_limit` to "limited").

### C2 — pre-keygen orphan guard fails open (`setup.rs`)
**What:** both reads in `handle_orphan_models_before_keygen` treated failure as "nothing to protect": cannot-open returned `Ok(())`, and `has_any_models().unwrap_or(false)` did the same for a query failure on an open store. Either way keygen proceeded — authorizing exactly the template-orphaning the guard exists to prevent. Both reads now abort with an actionable error. The enrollment marker is deliberately **not** consulted: this is a guard, not a reporter; a stale "not enrolled" marker must never authorize orphaning real templates.
**Tests (4):** unreadable store ⇒ error; open store + failing query ⇒ error (injected by repointing `face_models`' btree roots at orphaned wrong-type pages, which survives the migrations' unconditional `CREATE TABLE IF NOT EXISTS`); zero models ⇒ proceeds; models present ⇒ existing prompt/bail path.

### C3 — daemon auth swallows `list_models` (`handler.rs`, `direct.rs`)
**What:** see above. The same swallow shape survived both Wave-1 rewrites in `direct.rs` (`facelock test`'s backend, no rate-limit charge but the same falsehood) and now propagates with context there too.
**Tests:** `authenticate_storage_failure_is_error_and_charges_no_rate_limit` in `daemon_integration.rs`, modeled on the N11 exemption test's counter assertion.

### C4 — three failure policies for "does this user have models?" (`clear.rs`, `enroll.rs`, `test_cmd.rs`)
**What:** unified on **propagate**. The damaging one was `clear`'s D-Bus branch: a transient failure read as "No face models enrolled", printed success, and exited 0 having deleted nothing — while the direct branch on the identical failure assumed models existed and prompted for a doomed deletion. `enroll`'s stale-embedder check and `next_label` now propagate instead of silently skipping the warning or picking a wrong `-1` label; the post-success model-count advisory stays best-effort (a failed count must not turn a committed enrollment into a reported failure). No trait restructuring — per-site fixes only (Phase D owns the seam).
**Tests:** D-Bus reply interpretation (transport failure / daemon error reply / unexpected reply all propagate; model lists map to bool) and the direct branch against real unreadable and healthy stores; `next_label` failure propagation and suffix counting.

### C5 — three user-resolution implementations (`preview/mod.rs`, `status.rs`)
**What:** `preview` used getpwuid only, so `sudo facelock preview` (root-only since PR D) resolved to **root** and never recognized anyone; `status` omitted `DOAS_USER` and the getpwuid fallback. Both local implementations deleted; call sites use `ipc_client::resolve_user` (flag → `SUDO_USER` → `DOAS_USER` → `USER` → getpwuid).
**Test:** euid 0 cannot be obtained in a unit test, so the pin is structural (the local implementations no longer exist to diverge) plus a behavioral test of the shared resolver's `SUDO_USER` precedence and flag-wins-over-env — exactly the behavior the deleted getpwuid-only version lacked.

### C7 — `facelock test` states a falsehood (`test_cmd.rs`)
**What:** three-way discrimination. Store opens with zero models and store-absent-fresh keep today's "No face models enrolled" message; store present-but-unreadable is now an **error** naming the real problem, never "no models". For the message only, the per-user enrollment marker is consulted (readable in exactly the cases the database is not): "You appear to have N enrolled model(s), but the face database at … can't be read right now (…)." — hedged because the marker can be stale; it never influences the decision. Marker absent/unreadable (e.g. `--user` for a never-enrolled account) degrades to the plain accurate error.
**Tests (4):** all three branches plus the `--user other` degradation.
**Note:** the brief's literal instruction to fall back to "today's message" (the no-models text) when the marker is absent was not followed — with the store present but unreadable, that message is the exact falsehood C7 exists to kill. The fallback is the accurate plain error instead.

## Files touched

Owned: `crates/facelock-cli/src/commands/{setup,clear,test_cmd,status}.rs`, `crates/facelock-cli/src/commands/preview/mod.rs`, `crates/facelock-daemon/src/handler.rs`.
Narrow authorizations (declared): `crates/facelock-cli/src/commands/enroll.rs` (C4 unification), `crates/facelock-cli/src/direct.rs` (C3 same-shape), `crates/facelock-daemon/tests/daemon_integration.rs` + unit tests inside the edited CLI files (pinning tests), **plus** `crates/facelock-daemon/Cargo.toml` and `crates/facelock-cli/Cargo.toml` (+`Cargo.lock`): `rusqlite` as a **dev-dependency only**, same version/features as `facelock-store`'s, needed for the schema-level failure injection the C2/C3 tests require. Nothing new links into shipping binaries.

Not touched: `auth::pre_check`/`pre_check_audited` logic (called, not changed), `authenticate_with_embeddings`, the −1/−2/−3 sentinel encoding, `pam-facelock`, any D-Bus method signature, D-Bus policy, docs/book/man.

## PR-base chain

Based on `feat/cli-root-by-default` (PR D), which sits on `integration/wave-1`. Retarget to main as the underlying merges land.

## Cross-PR notes

- `remove.rs` root-before-prompt (A6) was verified already fixed by PR D and pinned by `cli_smoke.rs`; not redone here.
- A7's `legacy_database_present` clause is void — that helper was deleted in PR #111 (DEC-3).
- `bench.rs:679` (`current_user`) and `notifications.rs:54` (`original_user`) still carry local `SUDO_USER` chains. They were not named by A5 and serve different semantics (bench's display user, the notification target), but a future pass may want `bench.rs` on the shared resolver too — it omits `DOAS_USER`.
- `handle_authenticate` now checks `list_models` *before* acquiring the camera, so a storage-failed request no longer opens (or keeps warm) the camera. Behaviorally invisible to clients; noted in case a later PR reasons about camera acquisition ordering.

## Docs to reconcile (not edited here — includes PR D-owned `docs/contracts.md`)

- `docs/troubleshooting.md` / `book/src/troubleshooting.md`: `facelock test` can now report "You appear to have N enrolled model(s), but the face database can't be read…" for a present-but-unreadable store; the "No face models enrolled" text no longer appears for store failures anywhere (test/clear/enroll).
- `docs/contracts.md`: `setup` now refuses to generate an encryption key when the database cannot be read or queried (previously silently proceeded); `clear` exits non-zero on a failed pre-check instead of printing "No face models enrolled".
- Daemon `Authenticate` can return `Error {"storage error: …"}` where it previously returned a no-match `AuthResult`; PAM treats both as auth failure, but the audit/syslog surface differs (error vs. failed attempt) — worth a line in `docs/security.md`'s rate-limiting section: storage failures no longer consume rate-limit budget.

## Gates

- `just check` — clean (test, lint, fmt-check, audit; the two yanked-crate warnings are pre-existing and allowed).
- `just test-arch-layout` — 21/21.


tyvsmith:

PR #114, opened 2026-08-14
test: pin config-template, D-Bus wire, and log-filter conformance (E2/E3/E9)

tyvsmith:
## Summary

Three independent conformance tests, each replacing an assertion that either didn't exist or couldn't fail, per the [gap register](https://github.com/tyvsmith/facelock/blob/main/.omc/plans/gap-register.md) (this PR predates that file existing on `main`; see the orchestrating agent's task for context).

### E2 — config template ↔ `Config::default()`

`crates/facelock-core/tests/config_template_test.rs` parses the shipped `config/facelock.toml` (which ships byte-for-byte as `/etc/facelock/config.toml` on every install, per `%config(noreplace)`) and asserts it equals `Config::default()`. Since `Config` has neither `PartialEq` nor `Default` (and adding either touches non-test code in `facelock-core`, outside this PR's file ownership), the reference default comes from `Config::parse("")` — every field is `#[serde(default)]`, so an empty document is defaults by construction — and comparison goes through `Debug` formatting.

A second test uncomments every commented-out example value in the template (tracking `[section]` context so a key like `path` under both `[device]` and `[audit]` is handled per-section) and asserts that round-trips to the same default too. Three examples are deliberately excluded because they are illustrative, not defaults: `device.path` (shows syntax for an explicit camera path; the real default is auto-detect/`None`) and `recognition.detector_sha256`/`embedder_sha256` (ship the literal placeholder `"..."`, which `Config::validate()` rejects as not-hex). A fourth test makes that exclusion a live assertion instead of a silent skip.

**Drift found and fixed (template side):** `notification.notify_on_failure` had a real, live divergence. Its per-field serde default (`default_true`) disagrees with `NotificationConfig`'s manual `Default` impl (`false`) whenever the `[notification]` header is present but the key itself is omitted — which is exactly the shape of the shipped template (and of any minimal custom `[notification]` section a user writes, e.g. just `mode = "off"`). The doc comment already said "Default: false"; the field-level parser default silently produced `true`. **Fixed on the template side** by promoting `notify_on_failure = false` from a commented example to an active line, closing the gap for every fresh install without touching non-test code. **Not fixed:** the underlying field-level default in `crates/facelock-core/src/config.rs` (`notify_on_failure`'s `#[serde(default = "default_true")]` should be `default_false`/unadorned) is still live for any user config that specifies `[notification]` without this key — see "Docs to reconcile."

`[security.pam_policy]` is real syntax (documented in the template, parsed by `pam-facelock`'s own `PamConfig`) but has no corresponding field in `facelock-core::Config`. Confirmed by design, not drift: `pam_policy` is a PAM-only concern (allow/deny PAM service names), and `Config::parse` silently ignores unknown keys (no `deny_unknown_fields`), which is the correct behavior. Pinned as its own test.

Also added, per `domain-object-map.md` §1: a `#[cfg(test)]`-only test in `crates/pam-facelock/src/lib.rs` (`shipped_config_template_parses_as_pam_config`) that `include_str!`s the same shipped template through `PamConfig`, so a future key one schema accepts and the other rejects becomes a visible CI failure instead of a silent parse error on a real system.

### E3 — exact D-Bus signature strings

`crates/facelock-core/tests/dbus_types_test.rs` previously asserted `!sig.to_string().is_empty()` for every D-Bus type — true for any struct, so it caught zero regressions. Replaced with the exact zvariant signature for each type crossing the bus:

| Type | Signature |
|---|---|
| `AuthResult` | `(bisd)` |
| `ModelInfo` | `(usstss)` |
| `PreviewFaceInfo` | `(ddddddb)` |
| `DeviceInfo` | `(sssb)` |

Kept the `BUS_NAME`/`OBJECT_PATH`/`INTERFACE_NAME` constant pins and added one more: that `BUS_NAME` and `INTERFACE_NAME` currently coincide, pinned independently of their literal values. No signature was changed — only pinned as observed today, per the assignment's constraint.

### E9 — one shared `EnvFilter` default

Found three `tracing_subscriber` init sites in `facelock-cli` plus one in `facelock-bench`:
- `crates/facelock-cli/src/main.rs:244` — `EnvFilter::from_default_env()` with **no fallback at all** (a milder cousin of the historical bug: silently shows nothing without `RUST_LOG` set, rather than filtering on the wrong crate name).
- `crates/facelock-cli/src/commands/auth.rs:32-44` — already fixed in Wave 1 to `"facelock=info,facelock_daemon=info"`, with a comment explaining the `facelock` vs `facelock_cli` bug.
- `crates/facelock-cli/src/commands/daemon.rs:1226-1232` — already used the correct targets but as its own literal string (`"facelock_daemon=info,facelock=info"` — same targets, different order, no shared source of truth).
- `crates/facelock-bench/src/main.rs:57` — bare `tracing_subscriber::fmt::init()`, no explicit fallback.

Added `crates/facelock-cli/src/logging.rs`: `DEFAULT_LOG_FILTER = "facelock=info,facelock_daemon=info"` and `default_env_filter()`, used at all three `facelock-cli` sites. Its test module pins the constant's content and target names, and walks this crate's `src/` tree at test time to assert no other file builds an `EnvFilter` directly via `EnvFilter::try_from_default_env()`/`from_default_env()`, and that `main.rs`/`auth.rs`/`daemon.rs` each call the shared helper. **Honest scope, stated in the test's own comment:** this is a textual scan, not a borrow-checked guarantee — it catches the actual historical failure mode (copy-pasting the old hand-rolled block into a new command), not every conceivable bypass.

`facelock-bench` cannot import `facelock-cli`'s constant — `facelock-cli` is bin-only with no lib target (gap D6), and this PR's file ownership doesn't extend into `facelock-core`, the one crate both already depend on where a truly shared constant could live. Gave it an independently-pinned analog instead: `DEFAULT_LOG_FILTER = "facelock_bench=info"` (its bin target name already equals its crate name, so it never had the crate-vs-bin mismatch bug — this is purely about giving it an explicit, pinned fallback for consistency) plus the same style of scan test, adapted for its single-file layout (the scan stops at the `#[cfg(test)] mod tests` boundary so the test's own string literals don't trip its own check).

## Cross-PR notes

- No auth-semantics changes. No D-Bus signature changes (E3 pins today's wire format; nothing was retyped/reordered).
- `notify_on_failure`'s field-level default bug in `crates/facelock-core/src/config.rs` (see E2 above) is real and still live for user-written configs, just no longer live in the shipped template. Fixing it is a one-line change (`#[serde(default = "default_true")]` → `#[serde(default)]` or `default_false`) but touches non-test code in `facelock-core`, which was out of this PR's file ownership.
- `-p pam-facelock` fails to build standalone (`zbus` feature-resolution error: `default-features = false, features = ["blocking"]` alone doesn't pull in an async backend) — confirmed pre-existing on `integration/wave-2` before this branch's changes, unrelated to this PR. All tests here were verified via full-workspace `cargo test`/`just check`, where feature unification resolves it.
- The E9 fix surfaces (but does not act on) the D6 gap again: `facelock-cli` having no lib target blocks not just integration-testing the D-Bus authz layer, but also sharing a simple logging constant with `facelock-bench` without either duplicating it or reaching into `facelock-core`.

## Docs to reconcile

- `docs/contracts.md` (or wherever the D-Bus wire format is documented, if anywhere) should reference `dbus_types_test.rs` as the source of truth for current signatures.
- Fast follow-up: fix `notify_on_failure`'s field-level default in `crates/facelock-core/src/config.rs` (see above).
- Fast follow-up: once D6 (facelock-cli lib target) or a facelock-core edit is in scope, move `DEFAULT_LOG_FILTER`/`default_env_filter` to a single shared location for both `facelock-cli` and `facelock-bench`.

## Gate output

`just check` (test + lint + fmt-check + audit) passes clean on the full workspace. Tail:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
cargo fmt --all -- --check
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 1216 security advisories (from /home/ty/.cargo/advisory-db)
Crate:     core2       Version: 0.4.0   Warning: yanked   (pre-existing, unrelated)
Crate:     uds_windows  Version: 1.2.0   Warning: yanked   (pre-existing, unrelated)
warning: 2 allowed warnings found
```

Full-workspace `cargo test --workspace` shows every test binary green, including the new/changed ones:
```
tests/config_template_test.rs  ... 4 passed; 0 failed
tests/dbus_types_test.rs       ... 6 passed; 0 failed
facelock (bin, facelock-cli)   ... 236 passed; 0 failed   (includes 3 new logging:: tests)
facelock_bench (bin)           ... 10 passed; 0 failed    (includes 2 new logging tests)
pam_facelock (lib)             ... 27 passed; 0 failed    (includes shipped_config_template_parses_as_pam_config)
```

Not run here per instructions (shared podman image tag with concurrent agents): `just test-arch-pam`, `just test-arch-layout`. The orchestrator runs these centrally before the PR leaves draft.

### DIFF
--- a/config/facelock.toml
+++ b/config/facelock.toml
@@ -329,8 +329,12 @@
 # notify_on_success = true
 
 # Show notification on failed face match.
-# Default: false (failure is already visible — you get a password prompt)
-# notify_on_failure = false
+# Default: false (failure is already visible — you get a password prompt).
+# Set explicitly here, unlike the options above: this key's per-field parser
+# default disagrees with the type's overall default when the [notification]
+# header is present but the key itself is omitted, so a fresh install must
+# ship the value to get the documented behavior.
+notify_on_failure = false
 
 
 # ── Audit ─────────────────────────────────────────────────────────
--- a/crates/facelock-bench/src/main.rs
+++ b/crates/facelock-bench/src/main.rs
@@ -25,6 +25,26 @@ const WARM_ITERATIONS: u32 = 10;
 /// Number of snapshots for enrollment benchmark.
 const ENROLLMENT_SNAPSHOTS: u32 = 5;
 
+/// Fallback `EnvFilter` directive used when `RUST_LOG` is unset (gap E9).
+///
+/// This crate's `[[bin]]` target and its Cargo.toml package name both are
+/// `facelock-bench` (target `facelock_bench`), so it does not have the
+/// facelock-cli's `facelock` vs `facelock_cli` naming mismatch. It also has
+/// no lib target to share a constant with facelock-cli's `logging` module
+/// across the crate boundary (see gap D6) and this crate's file ownership
+/// for this change does not extend into facelock-core, so this is an
+/// independently pinned analog rather than a truly shared constant — see
+/// `crates/facelock-cli/src/logging.rs` for the sibling definition.
+const DEFAULT_LOG_FILTER: &str = "facelock_bench=info";
+
+/// The `EnvFilter` this binary's tracing-subscriber init must use: honors
+/// `RUST_LOG` when set and valid, otherwise falls back to
+/// [`DEFAULT_LOG_FILTER`].
+fn default_env_filter() -> tracing_subscriber::EnvFilter {
+    tracing_subscriber::EnvFilter::try_from_default_env()
+        .unwrap_or_else(|_| DEFAULT_LOG_FILTER.into())
+}
+
 #[derive(Parser)]
 #[command(
     name = "facelock-bench",
@@ -54,7 +74,9 @@ enum Command {
 }
 
 fn main() -> Result<()> {
-    tracing_subscriber::fmt::init();
+    tracing_subscriber::fmt()
+        .with_env_filter(default_env_filter())
+        .init();
 
     let cli = Cli::parse();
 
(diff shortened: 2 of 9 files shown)


tyvsmith:

PR #115, opened 2026-08-14
refactor(store): store-local StoreError — failure class as a type (D2)

tyvsmith:
Wave 3, PR H1 — the Store domain (gap D2, closing the type half of D3). Base: `integration/wave-2`.

## What this is

`facelock-store` could not say *why* it failed: every failure was collapsed at one line (`db.rs:17-19`, `fn map_err(e) -> FacelockError { FacelockError::Storage(e.to_string()) }`) into a stringly core-crate variant that is matched on zero times anywhere. Wave 2 (#113) fixed the *policies* behaviorally — guards fail closed, reporters propagate — but had to encode them with booleans, message strings, and a create-based open that could never observe "absent" (it papered over absence by materializing an empty database). This PR gives those policies the type they need so they can't regress. **Representational, not behavioral**: no auth-semantics change, no D-Bus wire change, `pam-facelock` untouched (it never links the store).

## The variant set

`StoreError` lives **in `facelock-store`** (`crates/facelock-store/src/error.rs`) — the store adds failure classes without editing the crate everything depends on. The crate's public API now returns `Result<T, StoreError>`; it no longer uses `facelock_core::error::Result`.

| Variant | Maps from | Meaning |
|---|---|---|
| `Absent { path }` | `ENOENT` on the pre-open stat (`open_existing` / `open_readonly`) | A fresh install — a *state* a caller may legitimately proceed on. Never returned by `create`. |
| `Denied { path, detail }` | `SQLITE_CANTOPEN`, `SQLITE_PERM`, `SQLITE_READONLY` (extended codes incl. `READONLY_CANTINIT` fold into their primary); any non-ENOENT stat error; a failed `chmod 600` on db/WAL/SHM | Present but inaccessible — "cannot tell", nothing may be concluded about contents |
| `Busy { path, detail }` | `SQLITE_BUSY`, `SQLITE_LOCKED` | Locked by another connection; transient |
| `Corrupt { path, detail }` | `SQLITE_NOTADB`, `SQLITE_CORRUPT`; a non-regular-file at the path; wrong-size template blobs | Present but not usable as a database |
| `Migration { path, detail }` | any failure inside `run_migrations` | Schema can't be brought forward |
| `Query { path, detail }` | everything else (constraints, I/O errors, not-found rows) | Residual class no caller policy distinguishes |

One `classify()` replaces `map_err`. A deliberate refinement over a bare `is_file()` check: **only ENOENT is evidence for `Absent`** — an unstatable path (e.g. unsearchable parent dir) is `Denied`, because a guard reading "cannot stat" as "absent" would be a fail-open. `SQLITE_CANTOPEN` also maps to `Denied`, not `Absent`: by the time SQLite reports it, existence was already checked, so it covers the stat-then-open race in the fail-closed direction.

`From<StoreError> for FacelockError` preserves the full rendered message, so **zero call sites needed edits to compile** — the workspace built unmodified after the retype. `database_exists` stays (its doc now points callers who need absent-vs-unstatable at the variants).

## Per-call-site policy (variant → behavior)

New `direct::open_store_existing(config) -> Result<FaceStore, StoreError>` (same layout hook as `open_store`, never creates) serves the sites below. `open_store` (create-based) remains for the flows that legitimately create: `direct::enroll`, direct/oneshot auth (the SQLite-backed rate limiter must be able to bring its storage into being — per #113, switching auth to `open_existing` would change auth-path behavior).

| Site | Shape | `Absent` | `Denied`/`Corrupt`/`Busy`/`Migration`/`Query` |
|---|---|---|---|
| `setup.rs` pre-keygen orphan guard (C2) | guard | **proceed** — no database file ⇒ provably nothing to orphan (and the probe no longer creates an empty DB to find that out) | **bail** — "refusing to generate a new encryption key…" (unchanged wording) |
| `test_cmd.rs` `direct_user_has_models` (C7) | reporter | "no models enrolled" (truthful, nothing created) | marker-aware error ("you appear to have N enrolled model(s), but…"), never "no models" |
| `test_cmd.rs` stale-embedder recheck | reporter | store proved present one query ago ⇒ unreachable; if it vanished mid-command, error instead of re-creating | propagate |
| `clear.rs` `direct_user_has_models` (C4) | reporter | "no models enrolled", exit 0 — without the old root-created empty DB side effect | propagate (never assume-yes, never "no models") |
| `clear.rs` delete path | actor | error (DB vanished after the probe) instead of re-creating an empty DB to delete from | propagate |
| `enroll.rs` `direct_has_stale_embedder` + `list_user_models`/`next_label` (C4) | reporters | "no models / first label" — enrollment *proper* remains the flow that creates the DB | propagate |
| `handler.rs` `handle_authenticate` (C3) | reporter | n/a (daemon holds an open store) | unchanged propagation; wire message loses the old `storage error: storage error: …` double prefix and now names class+path |

Behavioral deltas, all edge-case side effects and all in the safe direction (stated per the task's D3 requirement): (1) read-shaped probes no longer create an empty database at the path; (2) the orphan guard on a *missing parent directory* now proceeds (nothing can exist to orphan) where it previously bailed on the incidental create failure — with an unreadable/unstatable path it still bails.

## Proof the #113 pins hold

All pass with **assertions untouched**: `orphan_guard_tests::{unreadable_database_aborts_before_keygen, failing_models_query_on_open_store_aborts_before_keygen, zero_models_proceeds, models_present_non_interactive_bails_with_orphan_message}`, `test_cmd::tests::{absent_store_and_zero_models_read_as_not_enrolled, unreadable_store_is_an_error_not_no_models, unreadable_store_error_reports_marker_count, unreadable_store_for_other_user_degrades_to_plain_error}`, `clear::tests::*` (7), `enroll::tests::{next_label_propagates_store_failure, next_label_counts_existing_labels}`, and the C3 counter-unchanged test `authenticate_storage_failure_is_error_and_charges_no_rate_limit`. One justified non-assertion edit: a code comment inside `absent_store_and_zero_models_read_as_not_enrolled` said "created empty on first touch" — that side effect is deliberately gone, so the comment now describes the `Absent` value; the assertions are byte-identical.

New tests: store-level classification (`classify` code table incl. extended codes; `missing_file_is_absent_with_path`, `garbage_file_is_corrupt_not_absent`, `directory_at_path_is_corrupt_not_absent`, `locked_database_is_busy`, `unstatable_path_is_denied_not_absent`, `unreadable_file_is_denied_not_absent` — the last two skip under root, which bypasses permission bits); policy-level Absent ≠ Denied at the guard (`absent_database_proceeds_without_creating_one`) and reporters (`absent_store_probe_creates_nothing`, `direct_absent_store_reads_as_no_models_without_creating`, `absent_store_reads_as_no_models_without_creating`, `stale_embedder_check_propagates_unreadable_store`).

## Churn

13 files, +605/−123. Of that, `facelock-store` itself is +449/−97 (mostly tests and docs); the call-site commit is +156/−26 across 6 files. Compile-compatibility churn outside the store: **zero lines** (the `From` impl absorbed it).

## Gates

`just check` (test, clippy `-D warnings`, fmt-check, audit) — exit 0. Container tiers deliberately **not** run here (shared podman tag; orchestrator runs them centrally before this leaves draft).

## Cross-PR notes

- The conformance-tests agent (facelock-core / pam-facelock / EnvFilter sites): nothing here touches those files. If a conformance test asserts on store-failure *message text* over D-Bus, note the daemon's storage error messages changed shape: `storage error: <class-specific message incl. path>` instead of `storage error: storage error: <detail>`.
- P2-3 (`EnrollmentReport` / `DestructiveGuard` wrapper types) intentionally not built here — this PR provides the `StoreError` they consume; the guard/reporter split currently lives as per-site matches.
- `direct.rs` now has two store constructors; the future `FacelockBackend` trait (P2-5) should absorb both.

## Docs to reconcile

- `AGENTS.md` dependency table updated in-PR (`facelock-store` + thiserror); `CLAUDE.md` is a symlink, covered.
- `docs/contracts.md` / `book/src/contracts.md`: no store error contract is documented there today; nothing to update (binary names, paths, config keys, schema, auth semantics unchanged).
- `is_enrolled.rs` module doc extended to also forbid the new `open_store_existing` on the lock-screen hot path.

## Plan errata found

- `domain-object-map.md` §3 cites `maintenance.rs:36-39` as precedent — that file was deleted in Wave 1; the reasoning now lives in `FaceStore::open_existing`'s doc comment.
- The map's `Migration(..)`/`Query(..)` tuple sketch became struct variants carrying `path` — every variant can render a precise message without caller context.
- `architecture-remediation.md` §P2-2 says "keep `impl From<StoreError> for FacelockError` so nothing downstream breaks" — confirmed accurate; it was sufficient for the entire workspace.
- §P2-1's claim that `status`/`tpm` want `open_existing`: they use `open_readonly`, which now also returns `Absent`/`Denied`/… (same stat classification), so the claim is satisfied without changing those call sites.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -79,7 +79,7 @@ Supported providers: `cpu` (default), `cuda` (NVIDIA), `rocm` (AMD), `openvino` 
 | facelock-core | serde, toml, thiserror, tracing, subtle, zeroize, zvariant |
 | facelock-camera | facelock-core, v4l, image, tracing |
 | facelock-face | facelock-core, ort, ndarray, image, sha2 |
-| facelock-store | facelock-core, rusqlite (bundled), bytemuck |
+| facelock-store | facelock-core, rusqlite (bundled), bytemuck, thiserror |
 | facelock-daemon | facelock-core, facelock-camera, facelock-face, facelock-store, facelock-tpm, signal-hook, nix |
 | facelock-cli | facelock-core + all above, clap, reqwest, zbus, tokio, dialoguer |
 | facelock-bench | facelock-core, facelock-camera, facelock-face, facelock-store, clap, anyhow, tracing |
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1157,6 +1157,7 @@ dependencies = [
  "facelock-core",
  "rusqlite",
  "tempfile",
+ "thiserror",
 ]
 
 [[package]]
--- a/crates/facelock-cli/src/commands/clear.rs
+++ b/crates/facelock-cli/src/commands/clear.rs
@@ -2,6 +2,7 @@ use anyhow::Context;
 
 use facelock_core::Config;
 use facelock_core::ipc::{DaemonRequest, DaemonResponse};
+use facelock_store::StoreError;
 
 use crate::ipc_client;
 
@@ -37,7 +38,10 @@ pub fn run(user: Option<String>, yes: bool) -> anyhow::Result<()> {
     }
 
     if ipc_client::should_use_direct(&config) {
-        let store = crate::direct::open_store(&config)?;
+        // `open_store_existing`: has_models above just proved the database is
+        // there. If it vanished since, deleting has nothing to do — erroring
+        // beats re-creating an empty database in its place.
+        let store = crate::direct::open_store_existing(&config)?;
         let count = store
             .clear_user(&user)
             .map_err(|e| anyhow::anyhow!("{e}"))?;
@@ -66,9 +70,15 @@ pub fn run(user: Option<String>, yes: bool) -> anyhow::Result<()> {
 
 /// Direct-transport half of the pre-prompt check: a store that cannot be
 /// opened or queried is an error, never "no models" and never "assume yes and
-/// prompt anyway".
+/// prompt anyway". [`StoreError::Absent`] alone reads as "no models" — a
+/// fresh install has provably nothing to delete, and the probe must not
+/// create the database it is about to report empty.
 fn direct_user_has_models(config: &Config, user: &str) -> anyhow::Result<bool> {
-    let store = crate::direct::open_store(config)?;
+    let store = match crate::direct::open_store_existing(config) {
+        Ok(store) => store,
+        Err(StoreError::Absent { .. }) => return Ok(false),
+        Err(e) => return Err(e.into()),
+    };
     store
         .has_models(user)
         .map_err(|e| anyhow::anyhow!("storage error: {e}"))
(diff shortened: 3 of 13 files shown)


tyvsmith:

PR #116, opened 2026-08-14
refactor(config): parse once, resolve explicitly (D7)

tyvsmith:
Wave 3, PR H2 — the Config domain (gap D7): the parsed-vs-resolved split, per `domain-object-map.md` §1.

## Design

Two types in a new `crates/facelock-cli/src/resolved.rs`:

- **`ConfigLoad`** — *what the file says.* `main` performs the process's single read/parse and hands every command a `&Config`. Load-error rendering is unified in `ConfigLoad::require` (missing file → "run 'sudo facelock setup'"; broken file → path + parse error). `status` receives the whole `ConfigLoad` so a broken config is a finding it renders, not an exit.
- **`ResolvedConfig`** — *what is actually true.* Facts established by explicit probes, each tagged with a lightweight `Provenance` (`Probed`/`Claimed`) that H8's richer `Fact` type can adopt:
  - `ModelFiles` — configured detector/embedder presence (integrity stays with `FaceEngine`'s SHA256 check at load);
  - `CameraPresence` — configured device node exists; auto-detect stays a `Claimed` fact deferred to open time;
  - `ExecutionProviderFact` — is the configured EP compiled into the installed ONNX Runtime (`Available` / `NotBuiltIn` / `UnknownName` / `Unqueryable`); the decision rule is split from the ORT query so it tests without a GPU.

The `Config` type itself is untouched — the `#[serde(default)]` structure is the `%config(noreplace)` guarantee and stays as-is.

**Placement rationale:** `facelock-cli`, not `facelock-core`. The probes need `facelock-face` (ORT queries) — core cannot depend on it without inverting the crate DAG. The daemon's engine/store construction lives in `commands/daemon.rs` (in this crate), so daemon startup is covered; `facelock-daemon` (lib) receives already-built engine/store and needs no resolution machinery. The standalone `facelock-bench` binary keeps its own single top-of-process parse.

## Call-site accounting

At the branch point, facelock-cli had **27 `Config::load()` + 6 `Config::load_from` = 33 load expressions** in non-test code (the gap register's "29×" had drifted as Wave 2 landed — plan errata below).

After: **7 `Config::load()` + 6 `load_from` = 13**, every one deliberate and commented:

| Site | Count | Why it remains |
|---|---|---|
| `resolved.rs` (`ConfigLoad`) | 1 | **the** shared parse |
| `commands/daemon.rs` (`build_handler`) | 2 | daemon lifecycle: startup + mtime-triggered reload |
| `commands/auth.rs` | 2 | one-shot process spawned by PAM; its own top-of-process parse |
| `commands/setup.rs` | 4 | bootstrap: load-or-create-default (wizard + non-interactive entry points) |
| `commands/config.rs` | 3 | operates on the file itself (pre/post-edit validation) |
| `commands/enrollment_marker.rs` | 1 | `is-enrolled`'s tolerant read — must answer with no config |

Commands converted to take `&Config` from main: `enroll`, `remove`, `clear`, `list`, `test`, `preview`, `devices`, `bench` (7 subcommands), `tpm` (5 subcommands + `reseal`), `encrypt`, `decrypt`, `audit`; `status` takes `ConfigLoad`. `setup::update_config_encryption_method` now takes `&Config` instead of re-reading (it was a mid-flow re-read on the `tpm seal-key`/`unseal-key` paths).

Also removed: the daemon's **per-request** notification-config re-read in the D-Bus `Authenticate` handler — `maybe_reload_handler` already refreshes the handler's config at method entry, so the handler's held config is used instead.

## Probe sites: converted vs deliberately left

| Site | Disposition |
|---|---|
| `enroll.rs` inline detector/embedder `.exists()` pair | **converted** → `ModelFiles::probe` |
| `test_cmd.rs` same pair (×2, incl. post-setup re-check) | **converted**; the re-probe after setup is deliberate (setup changed the disk) and commented |
| `status.rs` `check_models` | **converted** → `ModelFiles` |
| `status.rs` `check_camera` | **converted** → `CameraPresence` |
| `status.rs` `check_inference` | **converted** → `ExecutionProviderFact`; the old check stat-ed a private copy of the provider module's `.so` search paths (drift hazard) and said nothing about the *configured* provider |
| `daemon.rs` `build_handler` | **converted**: names missing model files before the engine's opaque load error; warns when the configured EP will silently fall back to CPU |
| `direct.rs` `resolve_camera_device` / `build_resolved_camera_device` | **left** — the camera domain's rich resolution (D8 owns it) |
| `auth.rs` oneshot device validation | **left** — PAM auth path; no resolution machinery there |
| `setup.rs` `check_model` + SHA256 | **left** — the download/repair flow's subject matter is presence *and* hash |
| `setup.rs` `resolve_execution_provider_auto` | **left** — wizard *selection* of a provider, not verification of a claim |
| `setup.rs` `warn_provider_preflight` (`/dev/nvidiactl`) | **left** — driver-node hardware hint, not an ORT-build fact |
| `setup.rs` `detect_tpm`, `tpm.rs`/`encrypt.rs` key/device path checks | **left** — encryption-domain flows, coherent in place |
| `status.rs` store/PAM/encryption checks | **left** — H8 (Health) turns status into a renderer over `Health` |

## Invariants preserved

- **`is-enrolled` never probes.** It is dispatched in `main` *before* `ConfigLoad::read`, keeps its tolerant `marker_dir_or_default()` load, and touches no bus/camera/store. Two structural pins in `resolved.rs` tests: a source-scan that fails if `is_enrolled.rs` names any resolution/transport/store token, and a `Config::load()` allowlist that fails (naming the file) if any command regrows a mid-flow re-read. These pin the module text, not the call graph — the layout container tier pins the runtime behavior (21/21 below).
- `hyprlock` and `config` display stay probe-free and unprivileged (dispatched before the shared parse).
- The PAM schema in `pam-facelock` is untouched (dependency ceiling; #114's conformance test covers drift).
- `setup.rs` line-oriented config editing (`set_config_scalar`, ~:1029) untouched — ugly and correct.
- **No backend/reachability work**: no bus probing added; `Provenance` and `ResolvedConfig` are documented as the slot where H6 adds a reachability fact.
- No auth-semantics or D-Bus wire changes: interface signatures untouched; `config/facelock.toml` untouched; nothing the #114 sibling pins is touched.

## Behavior notes (deliberate, small)

1. Config load errors now surface **before** the interactive sudo-escalation prompt (main parses first). Previously e.g. `facelock list` prompted for sudo and then failed on the broken config as root.
2. `facelock enroll` on a system with **no config file at all** now errors with the unified "run 'sudo facelock setup'" message instead of reaching the interactive setup-gate prompt. With a config present but setup incomplete (the packaged-install case), the interactive gate is unchanged.
3. Daemon notification-config freshness now rides the mtime reload (was: raw re-read per request, which also caught same-second edits mtime comparison misses). Documented at the site.
4. `facelock status`'s inference line now reports whether the configured EP is built into the installed ORT (this loads the ORT dylib during `status` unless the EP is `cpu`/unknown; failures are contained and rendered).
5. Daemon startup with missing model files exits with the files named instead of the engine's load error.

## Churn

Two commits, `facelock-cli` only: 18 files, +459/−209 (parse-once) and 11 files, +432/−63 (facts). No other crate touched.

## Tests

- `ConfigLoad`: missing-file and broken-file rendering; parsed-config-outlives-the-file (ownership demonstration).
- Facts: `ModelFiles` probe with temp dirs (per-file presence, `missing()` set, absent dir); `CameraPresence` claim-vs-probed distinction; `ExecutionProviderFact` decision rule with injected detection (unknown name and `cpu` provably never query ORT; GPU against built-in list; unqueryable runtime).
- Structural pins as described under invariants.

## Gates

- `just check` — clean (exit 0; workspace tests, clippy `-D warnings`, fmt-check, cargo-audit with the 2 pre-existing allowed yanked-crate warnings).
- `just test-arch-layout` — 21/21 (pins `is-enrolled` runtime behavior).

## Cross-PR notes

- **H6 (Backend):** add the daemon-reachability fact beside the existing three in `ResolvedConfig`; `Provenance::Probed` fits it as-is. `should_use_direct` remains the (unowned) reachability logic this PR deliberately did not touch.
- **H8 (Health):** `Provenance` is the adoption point — replace `Resolved<T>` with `Fact<T>` carrying `Unavailable { needs }` and re-render `status` over `Health`. The three fact types here are already constructed without root/camera/bus (models/camera facts are plain stats; EP fact's query is injectable).
- **#114 (conformance):** no template or schema changes here; `notify_on_failure` drift remains pinned on the template side.
- **Template**: no `config/facelock.toml` deltas needed by this PR.
- The sibling branch's `logging.rs` / `DEFAULT_LOG_FILTER` is not in this base; no logging changes made here beyond the daemon's existing filter.

## Docs to reconcile

- `CHANGELOG.md`: no entry added (the Wave-3 chain has not been adding per-PR entries; reconcile once the wave merges). Suggested text: parse-once config ownership + resolved facts, and the daemon notification-config sourcing note (behavior note 3).
- `docs/troubleshooting.md` / `book`: `facelock status`'s inference line wording changed (now reports EP-vs-ORT-build instead of `.so` presence paths).
- `man/facelock.1`: no flag changes; no update required.

## Plan errata

- Gap register D7 says `Config::load()` is called **29×**; measured at this branch point: 27 (`Config::load()`) + 6 (`load_from`) = 33 expressions. The counts drifted as Wave 2 landed.
- `domain-object-map.md` §1 "setup.rs:1029" — confirmed; `set_config_scalar` sits at that region and is untouched.
- The daemon's config-reload-on-mtime is at `commands/daemon.rs` `maybe_reload_handler` (~:643) — confirmed as the deliberate re-read the brief predicted.
- **Not in the plan:** the daemon also re-read the config from disk *per authenticated request* (notification section, `Authenticate` handler ~:732). Fixed here; worth knowing it existed when reviewing daemon config-freshness assumptions.
- `test_cmd`'s post-setup model re-check judges by the pre-setup config (setup may rewrite `recognition.*_model`); pre-existing behavior, preserved and now commented.

### DIFF
--- a/crates/facelock-cli/src/commands/audit.rs
+++ b/crates/facelock-cli/src/commands/audit.rs
@@ -9,7 +9,6 @@ pub fn run(config: &Config, tail: bool, lines: usize) -> Result<()> {
     // hard-errors rather than offering the usual sudo re-exec prompt.
     crate::ipc_client::require_root_scripted("sudo facelock audit")?;
 
-
     if !config.audit.enabled {
         println!("Audit logging is not enabled.");
         println!("Enable it in config:");
--- a/crates/facelock-cli/src/commands/audit.rs
+++ b/crates/facelock-cli/src/commands/audit.rs
@@ -4,12 +4,11 @@ use std::path::Path;
 use anyhow::{Context, Result};
 use facelock_core::config::Config;
 
-pub fn run(tail: bool, lines: usize) -> Result<()> {
+pub fn run(config: &Config, tail: bool, lines: usize) -> Result<()> {
     // DEC-6: `audit` is typically invoked scripted/non-interactively, so it
     // hard-errors rather than offering the usual sudo re-exec prompt.
     crate::ipc_client::require_root_scripted("sudo facelock audit")?;
 
-    let config = Config::load()?;
 
     if !config.audit.enabled {
         println!("Audit logging is not enabled.");
--- a/crates/facelock-cli/src/commands/auth.rs
+++ b/crates/facelock-cli/src/commands/auth.rs
@@ -17,6 +17,8 @@ use facelock_store::FaceStore;
 use tracing::{debug, error, info};
 
 pub fn run(user: String, config_path: Option<String>) -> i32 {
+    // Deliberate parse (D7): `facelock auth` is its own one-shot process
+    // spawned by PAM, so this is its top-of-process read — not a re-read.
     let config = match config_path {
         Some(ref p) => Config::load_from(Path::new(p)),
         None => Config::load(),
--- a/crates/facelock-cli/src/commands/bench.rs
+++ b/crates/facelock-cli/src/commands/bench.rs
@@ -206,7 +206,6 @@ fn cmd_warm_auth(config: &Config) -> Result<()> {
 }
 
 fn cmd_preview(config: &Config) -> Result<()> {
-
     println!("=== Preview Frame Benchmark ===");
     println!(
         "Measuring: frame capture + face detection ({} iterations)",
@@ -263,7 +262,6 @@ fn cmd_preview(config: &Config) -> Result<()> {
 }
 
 fn cmd_enrollment(config: &Config) -> Result<()> {
-
     println!("=== Enrollment Benchmark ===");
     println!(
         "Measuring: time to capture and embed {} snapshots",
@@ -310,7 +308,6 @@ fn cmd_enrollment(config: &Config) -> Result<()> {
 }
 
 fn cmd_model_load(config: &Config) -> Result<()> {
-
     println!("=== Model Load Benchmark ===");
     println!("Measuring: ONNX model load time (SCRFD + ArcFace)");
     println!();
--- a/crates/facelock-cli/src/commands/bench.rs
+++ b/crates/facelock-cli/src/commands/bench.rs
@@ -46,28 +46,21 @@ pub enum BenchCommand {
     Report,
 }
 
-pub fn run(command: BenchCommand) -> Result<()> {
+pub fn run(config: &Config, command: BenchCommand) -> Result<()> {
     // DEC-6: `bench` is root by default (direct-mode access needs the 0600
     // root:root database regardless of subcommand, and auth benchmarks may
     // need TPM access besides). Supersedes the old TPM-only conditional check.
     crate::ipc_client::require_root("sudo facelock bench <subcommand>")?;
 
     match command {
-        BenchCommand::ColdAuth => cmd_cold_auth(),
-        BenchCommand::WarmAuth => cmd_warm_auth(),
-        BenchCommand::Preview => cmd_preview(),
-        BenchCommand::Enrollment => cmd_enrollment(),
-        BenchCommand::ModelLoad => cmd_model_load(),
-        BenchCommand::Calibrate => cmd_calibrate(),
-        BenchCommand::Report => cmd_report(),
+        BenchCommand::ColdAuth => cmd_cold_auth(config),
+        BenchCommand::WarmAuth => cmd_warm_auth(config),
+        BenchCommand::Preview => cmd_preview(config),
+        BenchCommand::Enrollment => cmd_enrollment(config),
+        BenchCommand::ModelLoad => cmd_model_load(config),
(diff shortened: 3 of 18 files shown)


tyvsmith:

PR #118, opened 2026-08-14
refactor(notify): Notifier seam + unified notification defaults (D9)

tyvsmith:
Wave 3, PR H4 — Notifications domain ([domain-object-map §6](../blob/main/.omc/plans/domain-object-map.md), gap **D9**). Base: `refactor/camera-caps` (H3).

## NotifyEvent vocabulary shipped

```rust
pub enum NotifyEvent {
    Scanning,
    Success { label: Option<String>, similarity: f32 },
    Failure { reason: String },
}
pub trait Notifier { fn notify(&self, event: &NotifyEvent); }
```

Enumerated from what the code actually sends today — these three, nothing else. **The map's mention of "enroll progress" events found no basis in code**: no enroll path sends any notification, so no enroll variants shipped (adding them would be new behavior, not refactor). Events carry data only; rendering (`body`/`icon`/`timeout_ms`) happens at the implementation, which is also where the D10 i18n boundary will want it.

## Trait placement

`facelock-core/src/notify.rs`: `NotifyEvent`, `Notifier`, `NullNotifier`, `NotifierFactory`, and the desktop-channel config policy (`should_notify_desktop` / `notify_desktop_if_enabled` — core already owns `NotificationConfig`). Zero new dependencies, and reachable from wherever H7 moves the D-Bus server. Implementations stay in `facelock-cli/src/notifications.rs` next to the setpriv machinery: `DesktopNotifier` (`for_current_session()` = session bus / SUDO_USER path; `for_user()` = daemon path), `TerminalNotifier` (renders bodies to stdout; unit-tested but not production-wired — terminal text in the auth path is the PAM conversation, which pam_facelock owns; carries `#[allow(dead_code)]` until H7/D10 wire it). `RecordingNotifier` lives in facelock-test-support.

The daemon learns its target user per request, so it is injected with a `NotifierFactory` (`Arc<dyn Fn(&str) -> Box<dyn Notifier>>`) rather than a single notifier — built by `notifications::daemon_notifier_factory()` and passed by `main` into `commands::daemon::run()`.

## Drift fix 1 — core `notify_on_failure` (with history findings)

`NotificationConfig::default()` says `false`; the field said `#[serde(default = "default_true")]`. **History: both sides shipped in the initial commit (`04ec3ef`) already disagreeing** — this was never a regression, just an undetected birth defect. Intent is unambiguous: the struct `Default`, the shipped template ("Default: false (failure is already visible — you get a password prompt)"), and `docs/configuration.md:118` all say `false`. Fixed the serde side to `#[serde(default)]`.

Swept the rest of `NotificationConfig`: `mode`, `notify_prompt`, `notify_on_success` were already consistent. Pinned: `config.rs` test `notification_section_present_with_keys_omitted_equals_default` (present-section/keys-omitted ≡ absent-section, field by field).

Sanctioned behavior change, for the record: a user who set `mode = "desktop"|"both"` but omitted `notify_on_failure` stops getting failure popups — which is what every piece of documentation always promised. The shipped template's active-header/commented-keys shape previously produced `true` here; note the template pin in the sibling PR #114 is *not* in this branch's base, and this PR deliberately doesn't touch the template.

## Drift fix 2 — PAM `PamNotificationMode`

Enum derived `#[default] Both`; struct `Default` says `Terminal`; benign only because the sole accessor `terminal()` treats both identically. Moved `#[default]` to `Terminal` (agreeing with the struct and with core's canonical `NotificationMode`). Ceiling respected — no facelock-core dep; alignment is a cross-referencing comment on `PamNotificationConfig` naming the canonical schema and the NotifyEvent mapping (`notify_prompt` ↔ Scanning, `notify_on_success` ↔ Success), plus two tests:

- `notification_section_present_with_keys_omitted_equals_absent` — same pin as core's, field by field.
- `shipped_template_notification_settings_equal_defaults` — `include_str!`s the same `config/facelock.toml` core ships and requires PAM's effective notification settings to equal its defaults. The template's active `[notification]` header with all keys commented is exactly the drift-triggering shape, so this is a live conformance check, not a tautology.

## Call sites converted

- `commands/daemon.rs` — **zero mentions of `crate::notifications` remain** (the D9 blocker metric). `FacelockService` holds the injected `notifier_factory`; the outcome→event decision is the pure `notify_auth_outcome()`.
- `commands/test_cmd.rs` — all 9 sites now `notify_desktop_if_enabled(config, &notifier, &event)` over a `DesktopNotifier::for_current_session()`.
- `main.rs` — injects `daemon_notifier_factory()` into `daemon::run()`.

Assertability evidence (`commands::daemon::tests`, via `RecordingNotifier`):
- `failed_auth_emits_failure_notification_when_enabled` — Failure event recorded;
- `failed_auth_emits_nothing_under_default_config` — same outcome, default (terminal-only) config, zero events;
- `successful_auth_emits_success_event_with_match_data` — label + similarity carried through.

## Delivery mechanics are verbatim (diff-level)

`original_user`, `send_notification_dbus`, `send_notification`, `send_as_user` moved behind the trait with bodies unchanged. Because `NotifyEvent` moved to core, its inherent render methods became module-local free functions; the **entire** diff of `send_as_user` (the setpriv path) is three call-syntax swaps:

```diff
-    let timeout = event.timeout_ms().to_string();
+    let timeout = timeout_ms(event).to_string();
-        event.icon(),
+        icon(event),
-        event.body().replace('\'', "'\\''"),
+        body(event).replace('\'', "'\\''"),
```

Same `setpriv --reuid … --regid … --init-groups --ambient-caps -all` invocation, same `runuser` fallback, same session-bus targeting, same filter-then-deliver order and debug logging. One unification: the CLI-path filter (`notify_if_enabled`) now emits the same debug-level "filtered by config" log the daemon path always had.

## Gates

- `just check` — clean (workspace tests, `clippy -D warnings`, `fmt --check`, audit with only the 2 pre-existing allowed yanked-crate warnings).
- `just test-arch-layout` — 21/21.

## Cross-PR notes (H7 especially)

- **H7**: `commands::daemon::run(config_path, notifier_factory)` — when the server moves out of facelock-cli, take `notify_auth_outcome` with it, keep receiving the factory from facelock-cli's `main`, and everything it needs is already in `facelock_core::notify`. The struct field is `FacelockService.notifier_factory`.
- **#114 (template pin)**: no conflict — this PR touches no template/packaging files; PAM's conformance test tolerates the active `notify_on_failure = false` line (unknown key to PAM, default-equal value to core).
- **D10 (i18n)**: rendering is now three free functions in one module — the future message-type seam has a single place to land.

## Docs to reconcile

None required: `docs/configuration.md`, `book/src/configuration.md`, and the template already document `notify_on_failure = false`; the code now agrees. Man pages don't state notification defaults.

## Plan errata

- Map §6's NotifyEvent sketch lists "enroll progress" — no such notification exists in code; vocabulary shipped without it (see above).
- `pam-facelock/src/lib.rs:939`'s comment attributes desktop notifications to "the daemon's auth_attempted D-Bus signal" — the daemon actually sends them directly via setpriv; the signal is a separate broadcast. Pre-existing comment inaccuracy, left for H7's sweep.
- `cargo test -p pam-facelock` alone fails to compile zbus (feature unification only provides zbus's tokio feature when facelock-cli is in the graph). Pre-existing; workspace-level tests (the `just check` gate) are unaffected.
- daemon.rs's capability-mask comments still say the drop exists for "`runuser`/`su`" while the code prefers setpriv — pre-existing wording, out of scope here.

### DIFF
--- a/crates/facelock-cli/src/commands/daemon.rs
+++ b/crates/facelock-cli/src/commands/daemon.rs
@@ -13,6 +13,7 @@ use facelock_core::dbus_interface::{
     AuthResult, BUS_NAME, DeviceInfo, ModelInfo, OBJECT_PATH, PreviewFaceInfo,
 };
 use facelock_core::ipc::{DaemonRequest, DaemonResponse};
+use facelock_core::notify::{Notifier, NotifierFactory, NotifyEvent, notify_desktop_if_enabled};
 use facelock_daemon::handler::Handler;
 use facelock_daemon::rate_limit::RateLimiter;
 use facelock_face::FaceEngine;
@@ -623,6 +624,10 @@ struct FacelockService {
     capture_slot: Arc<CaptureSlot>,
     /// Per-connection polkit frame-authorization cache (PreviewDetectFrame).
     preview_authz: Arc<PreviewAuthzCache>,
+    /// Builds per-user notifiers for auth outcomes. Injected from `main` so
+    /// the server never names the delivery implementation (D9) — a
+    /// prerequisite for moving this server out of facelock-cli.
+    notifier_factory: NotifierFactory,
 }
 
 impl FacelockService {
@@ -702,6 +707,7 @@ impl FacelockService {
         self.clear_camera_owner();
         let capture_guard = self.capture_slot.try_acquire("Authenticate")?;
         let handler = self.handler.clone();
+        let notifier_factory = self.notifier_factory.clone();
         let user = user.to_string();
         let signal_user = user.clone();
         let result = tokio::task::spawn_blocking(move || {
@@ -723,17 +729,7 @@ impl FacelockService {
             match response {
                 DaemonResponse::AuthResult(result) => {
                     // Send desktop notification (fire-and-forget, runs as root → setpriv)
-                    let event = if result.matched {
-                        crate::notifications::NotifyEvent::Success {
-                            label: result.label.clone(),
-                            similarity: result.similarity,
-                        }
-                    } else {
-                        crate::notifications::NotifyEvent::Failure {
-                            reason: "no match".into(),
-                        }
-                    };
-                    crate::notifications::notify_if_enabled_for_user(&notify_config, &event, &user);
+                    notify_auth_outcome(&notify_config, notifier_factory(&user).as_ref(), &result);
 
                     Ok(AuthResult {
                         matched: result.matched,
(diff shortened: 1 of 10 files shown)


tyvsmith:

PR #119, opened 2026-08-14
feat(i18n): message seam — gettext at the sink boundary (D10)

tyvsmith:
Implements the Messages domain (domain-object-map §7, gap D10) under the two decided constraints: **D1 — gettext everywhere** (libc FFI, no new dependencies) and **D2 — logs stay English**, with the boundary at the **sink**, not the string.

## The seam

`crates/facelock-cli/src/message.rs` — one module, three parts:

- **`UserMessage`** (~140 variants): a typed vocabulary of user-facing events carrying data, not strings — `EnrollComplete { model_id, count, label }`, not a formatted line.
- **Two renderers, one per sink class:**
  - `localized()` — gettext at the edge (domain `facelock`, `/usr/share/locale`, `FACELOCK_LOCALEDIR` override for testing). English fallback is **byte-identical** to the historical inline strings.
  - `machine()` — a C-locale event line derived from the variant's `Debug` form (`NoConfigFile { path: "/etc/..." }`). Never consults gettext.
- **Sinks:** `Terminal` (`info`/`error`/`prompt`/`confirm` — localized text to stdout/stderr, plus the machine line to `tracing::debug!` so `RUST_LOG=facelock=debug` shows the stable event stream beside what the user saw) and `fail()` (localized `anyhow` errors whose `Display` terminates on stderr).

The gettext wrapper is the same ~20-line libc FFI pattern `pam_facelock` uses; PAM keeps its own copy and its own text domain untouched (it cannot depend on anything here). The seam lives in `facelock-cli` rather than `facelock-core` because core's dependency contract excludes libc and no non-CLI crate renders user text today; when the daemon D-Bus server moves out of this crate (H7), the seam moves with it.

**D2 by construction:** a message routed through `UserMessage` cannot reach a log sink translated, because the machine rendering never touches gettext. Demonstrated live:

```
$ FACELOCK_LOCALEDIR=$PWD/target/locale LANGUAGE=xx RUST_LOG=facelock=debug facelock list --config /nonexistent.toml
DEBUG NoConfigFile { path: "/nonexistent.toml" }        <- trace: C locale
Error: XX-KEINE-KONFIG /nonexistent.toml                <- stderr: translated
```

`init()` sets **LC_MESSAGES only** (not LC_ALL), so LC_NUMERIC can never leak into in-process C libraries (ONNX Runtime parses floats), and pins the output charset to UTF-8.

## Conversion coverage (honest accounting)

Seam-routed rendering sites after this PR, per file:

| File | Seam-routed | Raw sites remaining | Remaining is |
|---|---|---|---|
| commands/setup.rs | 95 | 84 | deep sub-flows: TPM/encryption detail, orphan-model recovery, preset summaries, EP warnings, hyprlock handoff, PAM-extension hint, unit-install detail, helper bails |
| commands/test_cmd.rs | 16 | 5 | diagnostic bails (`daemon error:`, `unexpected response`), store-error wording |
| commands/enroll.rs | 12 | 5 | diagnostic bails |
| ipc_client.rs | 6 | 2 | `[y/N]` convention in `confirm()` (see residue), D-Bus call-context strings |
| commands/remove.rs | 5 | 1 | diagnostic bail |
| commands/clear.rs | 5 | 3 | diagnostic bails |
| resolved.rs | 2 | 0 | — (ConfigLoad::require fully converted) |
| notifications.rs | 1 | 1 | the `TerminalNotifier` println — it *is* the sink edge |
| main.rs | — | — | `message::init()` wired first |

**Unconverted this PR (the long tail, counted):** bench.rs 102, tpm.rs 49, hyprlock.rs 30, encrypt.rs 19, direct.rs 15, config.rs 13, devices.rs 11, list.rs 10 (partly `--json`-exempt), audit.rs 8 (renders machine records; stays English deliberately), status.rs 5, preview/* 9, state_layout.rs 1, auth.rs 1 (PAM-spawned; stderr goes to logs, not a human). Total raw sites went from ~566 to ~385; every conversion is mechanical per the pattern documented in `message.rs` module docs ("Adding a message").

What a non-developer sees running `setup` (full wizard spine: banner, all 9 step headers, prompts, selections, download flow, group membership, the PAM preview/confirm/backup/rollback block, summary, non-interactive outcomes), `enroll`, `test`, `remove`, `clear`, plus desktop/terminal notification bodies and the top-level config-load errors, is localizable.

## Machine-facing text that must NEVER localize (sink discipline)

- `--json` output anywhere (`list --json`, `is-enrolled --json`)
- `is-enrolled` stdout contract and exit codes (file untouched, structural pin still green)
- `audit.jsonl` and everything under `facelock audit`
- syslog and `tracing` output (daemon and oneshot)
- **D-Bus error message strings** — verified contract: PAM never shows daemon error text to users; `pam_code_for_daemon_error` (pam-facelock/src/lib.rs:864) substring-matches `"rate limited"` / `"IR camera required"` on the wire to pick PAM codes and syslog reasons. Those strings are wire protocol and stay English; PAM localizes its own presentation via its own gettext domain, untouched.
- anything container tests grep (`Matched model`, `No match`, …) — safe twice over: English fallback is byte-identical, and test containers run in the C locale.

## Toolchain (opt-in; gettext never required to build)

- `just pot` — regenerates both templates via `xgettext --language=C` keyed on the extraction keywords (`translate` for the CLI seam, `gettext` for PAM). xgettext has no Rust mode; C tokenization works because the seam keeps msgids as single-line plain literals (enforced by the documented pattern). Fails with a clear message if xgettext is absent.
- `just mo` — compiles `po/<lang>/{facelock,pam_facelock}.po` → `target/locale/<lang>/LC_MESSAGES/*.mo` (`msgfmt --check`); no `.po` files is a clean no-op.
- `just install` — ships `target/locale/` under `/usr/share/locale` **iff present**; `just uninstall` removes only our two domains. Nothing in build/test/CI invokes gettext tools.
- Manual verification recipe (also in the `mo` recipe comment): `msginit -i po/facelock.pot -o po/de/facelock.po -l de`, translate, `just mo`, then `FACELOCK_LOCALEDIR=$PWD/target/locale LANGUAGE=de facelock <cmd>`.

## .pot regeneration

- `po/facelock.pot`: **16 phantom entries** (strings existing nowhere in code) → **142 entries extracted from the seam**, every one backed by a `translate()` call with a source-location comment.
- `po/pam_facelock.pot`: 10 entries → **2** (`"Identifying face..."`, `"Face recognized."`). Plan errata: the domain map says PAM has 7 localized strings; the code wraps 4 call sites over 2 distinct msgids — the other 8 catalog entries were phantoms too.

## Tests

- Byte-identical English pins for representative messages (incl. container-grepped ones), a placeholder sweep across the whole vocabulary, NotifyEvent→UserMessage mapping, machine-rendering shape.
- Real catalog lookup: a minimal `.mo` built in-test (binary MO encoder), bound via `bindtextdomain` + `LANGUAGE` — asserts `localized()` translates, `machine()` does not, misses fall back; skips lookup assertions gracefully if the environment has no non-C locale.

## Residue / follow-ups

- `NotifyEvent::Failure { reason }` carries free text composed upstream — it interpolates untranslated inside a localized template. Structuring the reason is an H4 vocabulary follow-up.
- `ipc_client::confirm`'s `[y/N]` suffix and accepted answers (`y`/`yes`) stay English — answer parsing is a contract; localizing accepted answers needs a deliberate design (gettext's `rpmatch` pattern).
- The `Error:` prefix on fatal errors comes from `anyhow`'s termination path, and `Caused by:` chains stay English (deep error chains are long-tail by scope).
- Daemon-sent desktop notifications render in the daemon's locale (C under systemd → English), same as before this PR; per-recipient locale needs a user-locale lookup at notify time.
- dialoguer choice items (model preset names, PAM service descriptions) and summary sub-values (encryption/quality labels) are unconverted (counted above).
- Plural forms use the historical `model(s)` msgids (byte-identity constraint); moving to `ngettext` is mechanical later and PAM's ceiling allows it.

## Cross-PR notes

- **H4 (Notifier):** `NotifyEvent` untouched; its renderers (`body()`) now route through the seam. English rendering pinned by H4's existing tests, unchanged.
- **H2 (ConfigLoad):** `require()`'s two errors are seam events (`NoConfigFile`/`InvalidConfig`); H2's wording tests still pass on the byte-identical fallback.
- **H7 (daemon-server move):** the seam should migrate (to core or with the server) when the daemon leaves `facelock-cli`; `NotifierFactory` consumers get localization for free via `body()`.
- **Remediation §P3-2 (packaging conformance):** LC_MESSAGES install currently exists only in the justfile path; the six other packaging paths (deb/rpm/PKGBUILD/nix/openrc/tmpfiles) ship no catalogs yet — they need the same conditional install once translations exist.

## Docs to reconcile

- `docs/contracts.md`: note the two gettext domains (`facelock`, `pam_facelock`) as install artifacts and the D-Bus error strings as an English wire contract.
- `CONTRIBUTING.md` / `book`: a "translating facelock" section pointing at `just pot`/`just mo` and the message.rs pattern.
- `man/facelock.1`: `FACELOCK_LOCALEDIR` env var.

Gates: `just check` clean; `just test-arch-pam` 22/22; `just test-arch-layout` 21/21 (tails in the PR conversation below).

### DIFF
--- a/crates/facelock-cli/src/commands/clear.rs
+++ b/crates/facelock-cli/src/commands/clear.rs
@@ -3,6 +3,7 @@ use facelock_core::ipc::{DaemonRequest, DaemonResponse};
 use facelock_store::StoreError;
 
 use crate::ipc_client;
+use crate::message::{Terminal, UserMessage};
 
 pub fn run(config: &Config, user: Option<String>, yes: bool) -> anyhow::Result<()> {
     // ClearModels is root-only on the daemon side too, so demand root up front.
@@ -22,14 +23,14 @@ pub fn run(config: &Config, user: Option<String>, yes: bool) -> anyhow::Result<(
         daemon_user_has_models(ipc_client::send_request(&request))?
     };
     if !has_models {
-        println!("No face models enrolled for user '{user}'.");
+        Terminal.info(&UserMessage::NoModelsEnrolled { user: user.clone() });
         return Ok(());
     }
 
     if !yes {
-        let confirmed = ipc_client::confirm(&format!("Remove ALL face models for user '{user}'?"))?;
+        let confirmed = Terminal.confirm(&UserMessage::ConfirmClearAll { user: user.clone() })?;
         if !confirmed {
-            println!("Cancelled.");
+            Terminal.info(&UserMessage::Cancelled);
             return Ok(());
         }
     }
@@ -42,7 +43,10 @@ pub fn run(config: &Config, user: Option<String>, yes: bool) -> anyhow::Result<(
         let count = store
             .clear_user(&user)
             .map_err(|e| anyhow::anyhow!("{e}"))?;
-        println!("Removed {count} face model(s) for user '{user}'.");
+        Terminal.info(&UserMessage::ClearedModels {
+            count: count as usize,
+            user: user.clone(),
+        });
         // The user has no models by construction, so drop the marker outright.
         super::enrollment_marker::forget(config, &user);
         return Ok(());
@@ -54,7 +58,7 @@ pub fn run(config: &Config, user: Option<String>, yes: bool) -> anyhow::Result<(
 
     match response {
         DaemonResponse::Removed => {
-            println!("All face models removed for user '{user}'.");
+            Terminal.info(&UserMessage::AllModelsRemoved { user: user.clone() });
             super::enrollment_marker::forget(config, &user);
         }
         other => {
--- a/crates/facelock-cli/src/commands/enroll.rs
+++ b/crates/facelock-cli/src/commands/enroll.rs
@@ -6,6 +6,7 @@ use facelock_core::types::FaceModelInfo;
 use facelock_store::StoreError;
 
 use crate::ipc_client;
+use crate::message::{Terminal, UserMessage, fail};
 
 pub fn run(
     config: &Config,
@@ -20,16 +21,16 @@ pub fn run(
         let marker = std::path::Path::new(super::setup::SETUP_COMPLETE_MARKER);
         if !marker.exists() {
             ipc_client::require_root("sudo facelock setup")?;
-            println!("Setup has not been completed.");
-            if ipc_client::confirm("Run setup now?")? {
+            Terminal.info(&UserMessage::SetupNotCompleted);
+            if Terminal.confirm(&UserMessage::ConfirmRunSetupNow)? {
                 super::setup::run(false)?;
                 if !marker.exists() {
-                    anyhow::bail!("Setup did not complete successfully.");
+                    return Err(fail(UserMessage::SetupDidNotComplete));
                 }
                 // Setup includes face enrollment (Step 4), so we're done
                 return Ok(());
             } else {
-                println!("Run 'sudo facelock setup' when ready.");
+                Terminal.info(&UserMessage::RunSetupWhenReady);
                 return Ok(());
             }
         }
@@ -41,10 +42,7 @@ pub fn run(
     // warn prominently when the opt-in is active.
     if config.encryption.method == facelock_core::config::EncryptionMethod::None {
         if config.security.allow_plaintext {
-            eprintln!(
-                "WARNING: encryption.method = \"none\" and security.allow_plaintext = true.\n\
-                 Your face template will be stored UNENCRYPTED (plaintext biometric data at rest)."
-            );
+            Terminal.error(&UserMessage::PlaintextEnrollWarning);
         } else if let Err(message) = config.ensure_enroll_encryption_allowed() {
             anyhow::bail!(message);
         }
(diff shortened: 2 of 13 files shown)


tyvsmith:
Gate evidence (run locally, sole active agent on this podman host):

**`just check`** — exit 0 (workspace tests + `clippy -D warnings` + `fmt --check` + `cargo audit`):
```
cargo fmt --all -- --check
      Loaded 1216 security advisories
Crate:     core2      Version: 0.4.0   Warning: yanked
Crate:     uds_windows Version: 1.2.0  Warning: yanked
warning: 2 allowed warnings found        <- pre-existing, allowed
```

**`just test-arch-pam`**:
```

```

**`just test-arch-layout`**:
```

```

Runtime D2 demonstration (scratch `xx` catalog, since deleted):
```
$ FACELOCK_LOCALEDIR=$PWD/target/locale LANGUAGE=xx RUST_LOG=facelock=debug \
    ./target/debug/facelock list --config /nonexistent-i18n-test.toml
DEBUG NoConfigFile { path: "/nonexistent-i18n-test.toml" }
Error: XX-KEINE-KONFIG /nonexistent-i18n-test.toml
```

tyvsmith:

PR #120, opened 2026-08-14
refactor(backend): BackendKind selection + one fork per operation (D1, N10)

tyvsmith:
Wave 3, PR H6 — the Backend domain (gaps D1, N10, D3/D4 remnants): one question, one implementation, one failure policy. Based on `feat/i18n-seam` (full Wave-3 chain: H1 StoreError, H2 ConfigLoad/ResolvedConfig, H3 CameraCaps, H4 Notifier, H5 message seam).

## The seam

`crates/facelock-cli/src/backend.rs` (new): `Backend::select(config)` runs **once at the top of each backend-using command** — at most one deliberate, non-activating bus probe (`name_has_owner`, backend.rs:395) — and every operation that used to fork transport per call site forks in exactly one `Backend` method. A struct with per-op methods rather than a `Box<dyn>` trait: two implementations, crate-private, no object-safety or lifetime tax; the seam property (callers cannot reach around it) is held by the structural pins below.

- `BackendKind` (backend.rs:48), selection rule `classify` (backend.rs:342) — pure, probe injected, unit-tested without a bus
- `BackendCaps` (backend.rs:87) — capability facts stated at selection
- `DaemonReachability` (backend.rs:70) — the probe outcome recorded beside H2's resolved-config facts using `Resolved`/`Provenance` **as-is** (per H2's handoff note); `NotProbed` is a distinct value, never a guessed `Unreachable`

## BackendKind: three messages, three levels

| Kind | When | Log level | User-facing text |
|---|---|---|---|
| `Daemon` | mode=daemon, bus name owned | debug | none |
| `DirectByConfig` | mode=oneshot | debug | none — expected, not degraded |
| `DirectByFallback` | mode=daemon, no owner | **warn** | one stderr notice via the H5 seam (`DaemonUnreachableFallback`) |

The known misattribution is fixed: preview with a **stopped** daemon now renders `PreviewGraphicalDaemonUnreachable` instead of blaming "oneshot mode" as a configuration choice (was preview/mod.rs:22-30; the old message survives byte-identically for the genuine oneshot case). Pinned: `stopped_daemon_is_not_blamed_on_configuration` (preview/mod.rs).

## Per-site fork table (recount: 13 sites, plan said 12)

| Old fork site | Operation | Now |
|---|---|---|
| clear.rs:19 | has-models pre-prompt | `Backend::has_models` |
| clear.rs:38 | clear all | `Backend::clear_models` |
| enroll.rs:78 | stale-embedder check | `has_models` + `has_models_for_embedder` |
| enroll.rs:106 (via 149 helper) | list models (label scan) | `Backend::list_models` |
| enroll.rs:149 | enroll | `Backend::enroll` |
| devices.rs:10 | device listing | `Backend::list_devices` (one renderer; direct printing copy deleted) |
| preview/mod.rs:22 | preview routing | `caps().graphical_preview` + kind-accurate message |
| list.rs:32 | list models | `Backend::list_models` |
| remove.rs:27 | remove model | `Backend::remove_model` |
| test_cmd.rs:52 | has-models (C7) | `Backend::has_models` (+ marker-enriched wording kept in `enrolled_check`) |
| test_cmd.rs:76 | embedder match | `Backend::has_models_for_embedder` |
| test_cmd.rs:108 | test auth | `Backend::recognize` |
| enrollment_marker.rs:314 | marker recount | `Backend::list_models` via `refresh(&backend, ..)` |

After: **0** `should_use_direct` occurrences (the function is deleted); pinned by `the_boolean_fork_stays_deleted` and `daemon_transport_entry_points_are_single_sited` (backend.rs tests; `send_request`/`send_enroll` allowlist: the seam, ipc_client's definition, status.rs until H8, the preview frame loops). The is-enrolled probe-free pin in resolved.rs now names `Backend::select(`/`backend::` as forbidden tokens.

## D3 remnant: failure policy uniform per kind

Selection failure policy is decided once: `DirectByFallback` is chosen at selection and never re-probed — a daemon dying mid-invocation is an error, not a silent per-site re-fallback. Direct reads share one policy (`Backend::direct_read`): `StoreError::Absent` answers a benign value **without creating the database**; every other class propagates — never "no models" (C4/C7). This closes two residual creation side effects: `facelock list` and the marker recount no longer materialize an empty DB on a fresh install, and `remove` reports "not found" without creating one.

## N10: honest numbers

Re-measured on the base (the cited duplication had been rewritten twice since the plan):
- Still duplicated: the per-row decrypt loop (software-AAD unseal / TPM unseal / plaintext size-check+cast) — ~45 lines each in `Handler::load_user_embeddings` (handler.rs:224-316, 93 lines total) and `direct::load_user_embeddings` (direct.rs:249-320, 67 lines total), plus the fast-path condition.
- **Not** duplicated, left per-caller on purpose: sealer *initialization* policy (daemon fails enroll closed on a broken keyfile while auth continues; direct propagates) and the raw-row fetch/error-type wrapping.

Consolidated exactly the loop: `facelock_daemon::embeddings::decrypt_user_embeddings` (embeddings.rs:82) + `TpmAccess` (embeddings.rs:31, Held vs Lazy — the daemon holds a startup-initialized sealer, the one-shot path connects lazily) + `needs_raw_rows` (embeddings.rs:71). Both loaders are now fetch-and-delegate (handler.rs:226, direct.rs:254).

En-passant fix the unification forced: the handler previously took the **fast path** when `tpm.seal_database` was set but TPM init failed (or the `tpm` feature was absent), handing sealed blobs to `get_user_embeddings` → a misleading "corrupt store" error. Both callers now share the B3-correct condition; the handler's cfg-gated `tpm_sealer` field became an unconditional `Option<TpmSealer>` (passthrough reports a clear per-row "compile with tpm" error). No change for correctly-working setups; no pre_check/auth-semantics change; no wire change.

Pins that stayed green across both former call paths: direct.rs `load_user_embeddings_decrypts_software_sealed_rows`, `tpm_sealed_rows_error_clearly_without_tpm_support` (B3), `seal_database_slow_path_still_loads_plaintext_rows`; daemon integration suite (22 tests) incl. `keyfile_sealer_init_failure_fails_enroll_closed_no_plaintext`. New unit coverage in embeddings.rs (7 tests).

## D4 disposition

Closed, no code: `LAYOUT_IN_PROGRESS`/`StateReady` (and any thread-local re-entrancy guard) no longer exist anywhere in the tree — state_layout.rs was rewritten in Wave 1 (now 563 lines) and the cycle the guard protected against went with DEC-3's relocation removal.

## `enforces_auth_gates` decision

**Dropped** — it would state a falsehood: both transports run the same `pre_check` since Wave-2 N11. The surviving `facelock test` asymmetries are real but narrow — the direct path skips the SSH/lid physical-presence aborts (`PreCheckContext::test`) and audits as `Test` vs `Daemon`; neither path charges the rate limit for a failed test — and no caller branches on them, so they are documented at `Backend::recognize` (backend.rs) instead of carried as a boolean nobody reads. `BackendCaps` keeps the two facts callers actually consume: `graphical_preview` (daemon-only) and `device_formats` (direct-only; the D-Bus `DeviceInfo` carries none — wire unchanged).

## Behavioral deltas (beyond the spec'd messaging)

- `DirectByFallback` prints one stderr warning per invocation (was silent).
- Fresh-install direct reads no longer create the database (`list`, `remove`, marker recount).
- `devices` (direct, none found) gains the "Check that your camera is connected…" hint line the daemon path already printed; rendering otherwise byte-identical.
- `test` daemon path: an `Error` reply now surfaces as `daemon error: <msg>` (was `unexpected response from daemon: Error {..}`) and `Suppressed` renders as no-match (was "unexpected"); the desktop notification carries the error text — the direct path's existing policy.
- `test` renderings stay per-path byte-stable (daemon shows model id/label, direct does not) — presentation predating the seam, keyed on `kind()` with a comment; transport itself is single-sited in `recognize`.
- enroll's redundant in-branch `require_root` re-check removed (root already enforced at the top).
- Daemon corner: `seal_database` + failed TPM init now yields a clear per-row TPM error instead of a misleading "corrupt store" (see N10).

## Invariants held

PAM untouched and outside the trait; `DaemonRequest`↔dbus double translation untouched (H7's input type); is-enrolled/hyprlock/config never select or probe (pins updated, dispatch order unchanged); no D-Bus wire change; the PAM-matched protocol strings ("rate limited", "IR camera required") flow byte-identical (live in facelock-daemon/src/auth.rs, untouched); H5 discipline: the three new messages are human/stderr-only, machine sinks (`machine()`, D-Bus errors, audit) never localize.

## Cross-PR notes

- **H7 (daemon extraction):** the seam's daemon arms are now the only CLI consumers of `send_request`/`send_enroll` besides status.rs (Ping) and the preview frame loops — all named in the structural pin's allowlist. When the D-Bus server moves out and `DaemonRequest` becomes a handler input type, only backend.rs + those allowlisted files need touching; the pin will flag any strays.
- **H8 (Health):** `DaemonReachability` + `Resolved`/`Provenance` is the reachability fact ready to lift into `Fact<T>`; status.rs still runs its own Ping probe (deliberately left — H8's rewrite should consume `classify`'s fact and delete status.rs from the send_request allowlist).
- **i18n:** 3 new msgids regenerated into `po/facelock.pot` (`just pot`); translation catalogs need the usual update pass.

## Docs to reconcile

- `book/src/troubleshooting.md` / `docs/troubleshooting.md`: document the new one-line fallback warning ("daemon.mode = \"daemon\" but the facelock daemon is unreachable…") as the signature of a stopped daemon, and the corrected preview wording.
- `book/src/cli-reference.md`: `facelock devices` formats note (format detail requires direct camera access; the daemon listing has none) and the fresh-install `list` no-DB-creation behavior.
- No `docs/contracts.md` change needed: no binary/path/config/schema/wire/auth-semantics change.

## Gates

`just check` (test + clippy -D warnings + fmt-check + audit):
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.96s
cargo fmt --all -- --check
warning: 2 allowed warnings found   # pre-existing yanked-crate allowlist (core2, uds_windows)
```

`just test-arch-pam`:
```

```

`just test-arch-layout`:
```

```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1118,6 +1118,7 @@ dependencies = [
  "serde",
  "serde_json",
  "signal-hook",
+ "tempfile",
  "tracing",
  "tracing-subscriber",
 ]
--- a/crates/facelock-cli/src/backend.rs
+++ b/crates/facelock-cli/src/backend.rs
@@ -1,0 +1,732 @@
+//! Backend selection and the transport seam (D1).
+//!
+//! Every user-facing operation is answered either by the daemon over D-Bus or
+//! by direct (in-process) store/camera access. That used to be a boolean
+//! (`should_use_direct`) consulted independently at 13 call sites, each with
+//! its own failure policy and its own fresh bus probe — real TOCTOU windows
+//! (`clear` probed, prompted for an unbounded time, then probed again) and an
+//! activation asymmetry (`name_has_owner` does not start the daemon, but a
+//! `send_request` on the "wrong" branch would, silently flipping a later
+//! operation from direct to daemon mode — issue #89 validation fallout).
+//!
+//! Now selection is made **once per invocation**, at the top of each
+//! backend-using command, with **at most one** deliberate, non-activating bus
+//! probe. The outcome is a [`BackendKind`], not a boolean, because "direct"
+//! means two different things that deserve different messages and log levels;
+//! and every operation that used to fork per call site forks in exactly one
+//! method here.
+//!
+//! What stays outside this seam, on the merits:
+//! - **PAM** performs one operation and needs things the CLI must not have
+//!   (pinned-peer verification, a hard deadline, an env-cleared subprocess
+//!   fallback). It shares test vectors, not code.
+//! - **`is-enrolled`** is dispatched in `main` before any config parse and
+//!   must never probe the bus; `hyprlock` and `config` touch no backend.
+//! - **`status`** renders its own probes for now — H8 (Health) will consume
+//!   the [`DaemonReachability`] fact this module records.
+//! - Maintenance commands (`bench`, `encrypt`, `tpm`, `audit`, `auth`,
+//!   `daemon`) are direct-by-nature and never select.
+
+use anyhow::bail;
+use zbus::blocking::Connection;
+
+use facelock_core::Config;
+use facelock_core::config::DaemonMode;
+use facelock_core::dbus_interface::BUS_NAME;
+use facelock_core::ipc::{DaemonRequest, DaemonResponse, IpcDeviceInfo};
+use facelock_core::types::{FaceModelInfo, MatchResult};
+use facelock_store::StoreError;
+
+use crate::message::{Terminal, UserMessage};
+use crate::resolved::{Provenance, Resolved};
+use crate::{direct, ipc_client};
+
+/// Which implementation answers this invocation's questions — and, for the
+/// direct answers, *why* it is direct. The distinction carries the failure
+/// policy: an expected configuration is not a degraded state.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BackendKind {
+    /// `daemon.mode = "daemon"` and the bus name has an owner.
+    Daemon,
+    /// `daemon.mode = "oneshot"` — direct access is the configuration.
+    /// Expected; no warning.
+    DirectByConfig,
+    /// `daemon.mode = "daemon"` but the bus name has no owner — degraded.
+    /// WARNed once, at selection.
+    DirectByFallback,
+}
+
+impl BackendKind {
+    pub fn is_direct(self) -> bool {
+        !matches!(self, BackendKind::Daemon)
+    }
+}
+
+/// What the one probe observed, recorded beside the resolved-config facts
+/// (D7) with the same provenance discipline: `NotProbed` is a distinct value,
+/// never a guessed `Unreachable`. H8 (Health) lifts this into its `Fact`
+/// model; until then it is logged at selection.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DaemonReachability {
+    /// `mode = "oneshot"`: the bus is deliberately never asked.
+    NotProbed,
+    Reachable,
+    Unreachable,
+}
+
+/// What a backend cannot do, stated once at selection instead of discovered
+/// at the failure site.
+///
+/// Deliberately not carried here: `facelock test`'s remaining posture
+/// asymmetries. Both transports now run the same `pre_check` gates (N11), so
+/// the old `enforces_auth_gates` flag would state a falsehood; what still
+/// differs — the direct path skips the SSH/lid *physical-presence* aborts via
+/// `PreCheckContext::test` and stamps `AuditSource::Test` instead of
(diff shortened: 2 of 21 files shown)


tyvsmith:

PR #121, opened 2026-08-14
refactor(daemon): D-Bus server lives in facelock-daemon; handler input type internal (D5, D6)

tyvsmith:
Wave 3, PR H7 — closes gaps **D6** (the D-Bus server's authorization layer lived in a bin-only crate and could never be integration-tested) and **D5** (`DaemonRequest`/`DaemonResponse` leaked out of the handler into facelock-core and every CLI call site).

Zero wire change: bus name, object path, method names/signatures, the in-band `-2`/`-3` sentinel encoding, and the two PAM protocol strings (`"rate limited"`, `"IR camera required"`) are byte-identical. `dbus/org.facelock.Daemon.conf`, `dbus/org.facelock.Daemon.service`, the systemd unit, and `facelock-core/src/dbus_interface.rs` are untouched. `pam-facelock` is untouched except one comment correction; `is-enrolled` untouched.

## What moved (old → new)

| Was (`facelock-cli/src/commands/daemon.rs`, 2000 lines) | Now |
|---|---|
| `FacelockService` + `#[interface]` block | `facelock-daemon/src/server.rs:649` (generic service) / `:1145` (zbus glue on the production instantiation) |
| Authz table `Method`/`Scope`/`authorize_method` + unit tests | `server.rs:465` / `:533` / `:543` |
| `CallerIdentity` + `resolve_caller_identity` | `server.rs:426` / `:439` |
| Capture slot (DoS control) | `server.rs:104` |
| Polkit preview-frame gating (`PreviewAuthzCache`, `FrameAuthz`, watchers) | `server.rs:191`–`:420` |
| Idle timeout / `poll_shutdown`, sleep watcher, capability drop | `server.rs:1551` / `:1514` / `:1360` |
| Reload machinery (`maybe_reload_handler`) | `server.rs:716`, driven by an injected `HandlerRebuild` recipe |
| `notify_auth_outcome` (H4's decision fn) + its D9 tests | `server.rs:1290` |
| `run` / `run_dbus_server` | `server.rs:1310` / `:1429` |
| `build_handler` + camera factory + root check + tracing init | **stays in the CLI** — `facelock-cli/src/commands/daemon.rs` (222 lines), because handler construction leans on the CLI-shared `crate::resolved` (D7 probes) and `crate::state_layout` |

`ExecStart=/usr/bin/facelock daemon` and D-Bus activation are unchanged — `main.rs` still dispatches to `commands::daemon::run`, which now ends in `facelock_daemon::server::run(...)`.

## Minimal public surface shipped

- `facelock_daemon::server`: `run`, `ServerError`, `FacelockService<C, E>` + `::new`, the `*_as` method-level entry points, `CallerIdentity` (pub fields), `FrameAuthz`, `ProductionHandler`, `HandlerRebuild<C, E>`/`ProductionRebuild`. The service is generic over the handler's camera/engine; the `#[interface]` impl binds production types and owns only zbus glue (header→identity, polkit spawn, signal emission).
- `facelock_daemon::auth::AuthOutcome` / `facelock_daemon::enroll::EnrollOutcome` — the typed outcome vocabulary shared by both transports (same variant shapes the old enum had, so daemon-internal construction sites changed by name only; the handler maps them via `From`).
- Justification: the seam could have exported a request builder instead, but the CLI's two consumers (D-Bus client, direct path) both want *outcomes*, not requests — the request enum exists only to feed `Handler::handle`, which no CLI code does.

## D5 relocation

- `DaemonRequest`/`DaemonResponse` now live in `facelock-daemon/src/handler.rs:23`/`:55` ("a handler input type the CLI never sees"); `facelock-core/src/ipc.rs` keeps only the shared payload structs (`PreviewFace`, `IpcDeviceInfo`, `IpcFormatInfo`).
- The wire ↔ handler double translation **stays** (it is what keeps the handler transport-agnostic) — it is now invisible outside facelock-daemon.
- `ipc_client` grew one typed function per D-Bus method (`authenticate`, `list_models`, `remove_model`, `clear_models`, `list_devices`, `ping`, `preview_detect_frame`, `release_camera`; `send_enroll` now returns the tuple). Sentinel decoding and AccessDenied hinting unchanged.
- Two structural pins in `backend.rs`: the transport-entry-point allowlist now enumerates the typed functions per allowed file (replacing the `send_request`/`send_enroll` pin — updated deliberately per the H6 note), and a new pin fails if `DaemonRequest`/`DaemonResponse` ever reappear in facelock-cli sources.

## New integration coverage (`facelock-daemon/tests/server_authz.rs`, 9 tests)

Real `FacelockService` around a real `Handler<MockCamera, MockFaceEngine>` + in-memory store:

- every entry point except Authenticate denies a non-root caller (AccessDenied), including that a denied caller never starts a polkit round-trip (panicking check-starter);
- root passes authorization on every entry point (full success asserted where the mock flow guarantees it);
- Authenticate self-scope: alice→alice runs the real comparison, alice→bob denied, unresolvable username fails closed, root may act for anyone;
- similarity redaction: zeroed for non-root, real for root, outcome unchanged (N12);
- in-band encoding: `model_id == -2` with byte-exact `"rate limited"`, `-2` with `"IR camera required"`, `-3` for `suppress_unknown` — never a D-Bus error;
- N11: root's failed attempts don't charge the rate limit; the user's own do, and the next attempt is rejected in-band;
- root `PreviewDetectFrame` returns frames + unredacted scores with no polkit involvement.

**Honestly not covered here:** the zbus glue — caller-identity resolution from live message headers (`GetConnectionUnixUser`), signal emission, D-Bus activation, bus policy, real polkit round-trips, and the live config reload against `/etc`. That layer is exercised by the container tiers (real bus + PAM), which pass — tails below.

## ⚠️ Dependency-table change (AGENTS.md/CLAUDE.md)

`facelock-daemon` (a lib crate) now depends on **zbus (default-features off, `tokio` feature) + tokio** (plus futures-util, libc, thiserror). The AGENTS.md dependency table and the crate-purpose row are updated in the move commit. `facelock-cli` sheds its direct `tokio` and `futures-util` deps (zbus pulls its own runtime). E8 (facelock-daemon's `tracing-subscriber` dep) predates this PR and is unchanged — tracing *init* stays in the CLI.

## C8/C9 new homes (Phase E worker)

- **C8** (RemoveModel's unit wire reply loses "did it exist"): server mapping `crates/facelock-daemon/src/server.rs:924`; client `crates/facelock-cli/src/ipc_client.rs:274` (`remove_model`); seam surface `backend.rs::remove_model` (returns `None` on the daemon path).
- **C9** (devices format loss over the wire): server mapping `crates/facelock-daemon/src/server.rs:1074`; client reconstruction with `formats: vec![]` at `crates/facelock-cli/src/ipc_client.rs:327`.

## Behavior notes (reviewed intentionally, all non-wire)

- Dead code dropped, not relocated: `send_request`'s `Shutdown` arm (restart uses busctl) and `PreviewFrame` arm (no CLI caller); the `Ok(DaemonResponse::Error/Frame/other)` arms in both preview loops, status's "connected (unexpected response)" arm, and backend's `models_reply`/`removed_reply` error arms — all unreachable, because non-Authenticate daemon failures arrive as D-Bus errors (transport `Err`), which is exactly the C4 policy and now holds by construction. The four backend tests that pinned those dead arms went with them.
- Micro-reorder from the glue split: caller-identity resolution now happens *before* the `last_activity` store and `maybe_reload_handler` (was after). Observable only if identity resolution fails (no sender / bus proxy failure), in which case the idle clock and reload are now not touched — authorization decisions are unchanged.
- Live reload is disabled when no rebuild recipe is injected (tests); production always injects one, and `--config` runs keep reloading correctly because `Config::load()` honors the process-level override `main` sets.

## Cross-PR notes

- **H8 (Health):** `status.rs` now calls `ipc_client::ping()` (typed); the pin comment still marks it "until H8". `DaemonReachability` in `backend.rs` unchanged.
- The Wave-1 E3 conformance pins target `facelock-core/src/dbus_interface.rs`, untouched here.
- H4's errata is done: the PAM comment claiming notifications ride the `auth_attempted` signal is corrected (the daemon sends them directly via setpriv).

## Docs to reconcile (not in this PR)

`docs/security.md` / `docs/contracts.md` name `commands/daemon.rs` as the server's home in a few prose spots; the D-Bus *contract* content is unchanged. Worth a sweep when the wave merges.

## Gates

- `just check` (test + clippy `-D warnings` + fmt + audit): clean (two pre-existing allowed yanked-crate warnings).
- `just test-arch-pam` — **22/22** (tail):

```
TEST: env_clear: marker was active in the PAM process ... PASS
TEST: env_clear: LD_PRELOAD marker not loaded by oneshot child ... PASS
TEST: env_clear: SSH_CONNECTION survives to oneshot child ... PASS
TEST: env_clear: oneshot child PATH pinned to /usr/bin:/bin ... PASS

```
- `just test-arch-layout` — **21/21** (tail):

```
TEST: group member cannot read the database ... PASS
TEST: group member can read a model file by name ... PASS
TEST: non-member cannot traverse the state dir at all ... PASS
TEST: non-member cannot list models/ either ... PASS
TEST: non-member cannot read the audit log directory ... PASS

```


tyvsmith:

PR #123, opened 2026-08-14
fix(pam): build zbus with the feature set that actually compiles standalone

tyvsmith:
## What

Fixes the merge-blocker from REV-107's principal-architect review: `cargo build -p pam-facelock` did not compile on its own.

## Reproduction

```
$ cargo build -p pam-facelock
error: Either "async-io" (default) or "tokio" must be enabled
error[E0433]: cannot find module or crate `async_lock` ...
error[E0433]: cannot find module or crate `async_process` ...
error: could not compile `zbus` (lib) due to 21 previous errors
```

`crates/pam-facelock/Cargo.toml` declared:

```toml
zbus = { version = "5", default-features = false, features = ["blocking"] }
```

In zbus 5 the `zbus::blocking` module the PAM code uses (`zbus::blocking::Connection`, `::Proxy`, `::fdo::DBusProxy`, `::connection::Builder`) is gated by **`blocking-api`**, not `blocking`. `blocking` is only a sub-item of the async-io bundle that turns on an unrelated thread-pool crate. With `default-features = false` and no runtime selected, the feature set is incoherent and zbus cannot compile.

## Why it hid

`cargo build --workspace` unifies features across crates. `facelock-cli` and `facelock-polkit` both request `zbus = { ..., features = ["tokio", "blocking-api"] }` for the same zbus, so the workspace build silently repaired pam-facelock's set. `cargo tree` also resolved fine (58 crates) — graph resolution succeeds; it is not a compile check. Only a standalone `cargo build -p pam-facelock` exposed it.

This means the memory note claiming "pam solo-test breakage is pre-existing" was **wrong**. `main` declares `zbus = { version = "5", features = ["blocking"] }` (defaults *on*), which builds standalone via the async-io backend. This PR's commit `54a2935 fix(pam): build zbus without default features (C1)` added `default-features = false` and is what introduced the breakage.

## The fix

```toml
zbus = { version = "5", default-features = false, features = ["tokio", "blocking-api"] }
```

Matches `facelock-cli` and `facelock-polkit` exactly. pam-facelock now builds and tests in isolation (29 tests pass). The async-io thread-pool crates (`async-channel`, `async-task`, `blocking`, `piper`) drop out of `Cargo.lock`.

## Guard re-tuning (measured)

The CI guard forbade `async-io|async-signal|signal-hook-registry|polling`. The correct tokio backend pulls **`signal-hook-registry`** transitively via tokio's `process` feature, so the old guard would have rejected this very fix (it was calibrated against the broken tree).

Measured post-fix dependency set (`cargo tree -p pam-facelock --edges normal --prefix none`):
- **Added** by tokio backend: `bytes`, `errno`, `mio`, `signal-hook-registry`, `socket2`, `tokio`
- **Removed** async-io backend: `async-channel`, `async-task`, `atomic-waker`, `blocking`, `concurrent-queue`, `crossbeam-utils`, `piper`

New guard forbids the async-io runtime backend directly — `async-io`, `async-signal`, `polling`, plus `async-executor`, `async-fs`, `async-lock` — and drops `signal-hook-registry`. Verified: the new regex matches nothing in the fixed tree and still catches 5 markers in `main`'s async-io tree.

## Isolation build in CI

Added a `Build pam-facelock in isolation` step (`cargo build -p pam-facelock`, not `--workspace`) next to the guard, so this class of bug can no longer hide behind feature unification. Mirrored locally as `just check-pam-standalone`, wired into `just check`.

## Crate-count story (C1 goal holds)

| Tree | Crates | Compiles standalone? |
|------|--------|----------------------|
| pre-#107 baseline (`main`, async-io backend) | 68 | yes |
| #107 as-shipped (`default-features=false, ["blocking"]`) | 58 (resolves) | **no** |
| this fix (`["tokio","blocking-api"]`) | **57** | yes |

The fix is one crate smaller than the number #107 claimed (58) and 11 smaller than the 68-crate baseline — C1's dependency-shrink goal is preserved and slightly improved.

## Review finding

Fixes REV-107's merge-blocker: pam-facelock failing to build standalone.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,12 +51,32 @@ jobs:
       - name: Run tests
         run: cargo test --workspace
 
+      # Build the PAM module ALONE — not via --workspace. `cargo build --workspace`
+      # unifies features across crates, so facelock-cli/facelock-polkit (which both
+      # request zbus ["tokio","blocking-api"]) silently repair an incoherent zbus
+      # feature set in pam-facelock. That masks a real breakage: zbus's `blocking`
+      # module is gated by `blocking-api`, NOT by the `blocking` feature, and with
+      # default-features off you must also pick a runtime (`tokio` or `async-io`).
+      # A standalone build is the only thing that proves the PAM crate compiles on
+      # its own. Keep in sync with `just check-pam-standalone`.
+      - name: Build pam-facelock in isolation
+        run: cargo build -p pam-facelock
+
+      # The whole point of PR #107 is to keep the PAM module's dependency surface
+      # small, so forbid the async-io runtime backend: async-io/async-signal/polling
+      # plus the async-executor/async-fs/async-lock trio that only the async-io zbus
+      # backend pulls. We deliberately do NOT forbid signal-hook-registry: the
+      # correct tokio backend pulls it transitively via tokio's "process" feature,
+      # so listing it here would reject the very fix that shrinks the surface. If you
+      # think you need to add signal-hook-registry back, you have probably
+      # reintroduced the async-io backend — check for async-io first.
+      # Keep in sync with `just check-pam-standalone`.
       - name: Verify pam-facelock dependency surface
         run: |
           cargo tree -p pam-facelock --edges normal --prefix none | awk '{print $1}' | sort -u > /tmp/pam-deps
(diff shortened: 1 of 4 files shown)


tyvsmith:

PR #126, opened 2026-08-14
fix(security): always rate-limit Authenticate; give `facelock test` an explicit root-only method

tyvsmith:
## The vulnerability

`crates/facelock-daemon/src/server.rs` computed `let caller_is_root = caller.uid == 0;` and called
`handler.handle_authenticate(user, !caller_is_root)` — a root caller's failed authentication consumed
no rate-limit budget.

The intent (gap N11) was narrow: `facelock test` is root-only, and testing a few times shouldn't lock
you out of real auth. But **"caller is root" is not a proxy for "this is a test run"**:

- `pam_facelock` runs inside the PAM stack of the calling program. `sudo` is setuid-root; `login`,
  `su`, and root-run display-manager greeters run as root too. PAM tries D-Bus first, so a **real
  failed face authentication at a `sudo` prompt reached the daemon as caller UID 0 and was never
  charged.**
- Consequence: the documented core security rule (5 attempts/user/60s — `docs/security.md`,
  `docs/contracts.md`) was **inert on the project's primary documented PAM target**
  (`/etc/pam.d/sudo`). An attacker presenting spoof material at a sudo prompt had unlimited
  daemon-mediated attempts.
- Two independent signals that this was unintended: the oneshot fallback path *does* charge the same
  failure (`crates/facelock-cli/src/commands/auth.rs`), so whether an attempt counted depended purely
  on whether the daemon happened to be running; and `docs/contracts.md` justified the exemption with
  "Real end-user authentication (hyprlock, screen lockers) runs as the *unprivileged user
  themselves*, so it is always charged as before" — true for hyprlock, false for sudo/login.

Who is affected: anyone running the documented `auth sufficient pam_facelock.so` line in
`/etc/pam.d/sudo`, `/etc/pam.d/login`, `/etc/pam.d/su`, or a display-manager stack, with the daemon
running (the default `daemon.mode = "daemon"`). Screen lockers that run PAM as the user were always
charged correctly.

Root cause note: this shape came from the orchestrator's implementation instruction on PR #112, not
from the implementing agent's own design.

## The design: explicit beats inferred

Stop inferring intent from caller privilege. Make the test path a separate method.

1. **`Authenticate` always charges a failed attempt**, on every transport and for every caller
   including root. This restores the limiter on every PAM stack.
2. **New root-only `TestAuthenticate` D-Bus method** — same `s` argument, same `AuthResult` reply and
   the same `-1`/`-2`/`-3` sentinels. Two deliberate differences: it charges no rate-limit budget,
   and it applies `PreCheckContext::test()` (skips the SSH/lid physical-presence gates). Root-only is
   *why* a budget-free endpoint is safe to offer, so the scope is listed explicitly in the
   table-driven authorization rather than left to the `Scope::Root` catch-all.
3. **`AuthIntent`** (`facelock-daemon/src/handler.rs`) carries the purpose through the handler: what
   the attempt costs, which gates run, and which audit `source` it stamps. One table, one place.
4. `facelock test` calls the new method through the Backend seam. The direct/root path is unchanged
   and still never records failures structurally.

This also resolves the ssh/lid asymmetry PR #112 had to document as a wart: the daemon transport
previously had nowhere to apply the test context, so diagnosing with the lid closed only worked in
oneshot mode. Both transports now behave identically, and the daemon's diagnostic runs audit as
`test` instead of claiming the enforcement of an ordinary `daemon` authentication.

**PAM needs no change** — `crates/pam-facelock/src/lib.rs` calls `Authenticate`, which is now
correctly always-charging. The crate is untouched by this PR (`git diff --stat` confirms), as is
`is-enrolled`. The two protocol strings PAM substring-matches ("rate limited", "IR camera required")
still flow byte-identical through the same in-band encoding, pinned by the existing tests in
`tests/server_authz.rs`.

## Wire addition and every pin updated

- `dbus/org.facelock.Daemon.conf` — **no edit needed, verified**: the policy is interface-scoped
  (`send_interface="org.facelock.Daemon"` grants the whole interface to root and the `facelock`
  group); per-method decisions live in the daemon.
- `crates/facelock-core/tests/dbus_types_test.rs` — existing signature pins untouched; added a pin
  stating that `AuthResult` is now the reply of two methods and pinning the shared argument list.
  Method *names* cannot be pinned from core (zbus derives them from the `#[interface]` fn names), so
  the doc points at the daemon-side scan below.
- `crates/facelock-daemon/src/server.rs` — new `interface_methods_and_the_authz_matrix_are_the_same_set`
  scans the `#[interface]` block and requires it to match `Method::ALL` exactly. A wire method with
  no `Method` variant would be one the authorization matrix never sees. Falsified by removing the
  variant: the test fails.
- `crates/facelock-daemon/tests/server_authz.rs` — the matrix tests iterate `Method::ALL`, so
  `TestAuthenticate` is covered automatically (root allowed, non-root denied), plus explicit
  end-to-end denial tests.
- `crates/facelock-cli/src/backend.rs` — the single-sited transport pin now covers
  `test_authenticate`, and a new pin fails if any CLI source references `ipc_client::authenticate`
  (deleted: the CLI has no business calling the always-charging method).

## docs/contracts.md — the contradiction, fixed

The file contradicted itself about exactly this behavior. Both passages are reconciled with what
ships:

- "facelock test Semantics (N11)" (~:118) claimed consumption was exempted for root callers, with the
  hyprlock justification quoted above. Rewritten around the new method, including why UID is the
  wrong predicate.
- "Audit Log Entries" (~:284) claimed `test` against a running daemon "goes through `Authenticate`
  … and its failures count against the rate limit" — the opposite. Rewritten: `test` is the
  diagnostic path on either transport, `daemon`/`oneshot` are the enforced ones.
- Also updated in the same commit as the behavior: the method list, the authorization table, the
  capture-concurrency list, the `AuthAttempted` signal note, and the error-encoding section.

## Tests

Added:
- `root_failed_authenticate_charges_the_rate_limit` — **the regression pin.** Asserts the on-disk
  `rate_limit` table through a second connection, not just the next reply.
- `user_failed_authenticate_charges_the_rate_limit` — the unchanged half.
- `test_authenticate_does_not_charge_the_rate_limit` — three failed diagnostic runs against a budget
  of one, plus the proof that the rate-limit *check* is still not exempted.
- `test_authenticate_denies_a_non_root_caller` (own user and cross-user), and `TestAuthenticate` added
  to the sweep tests for both denial and root access.
- `tests/test_authenticate_ssh_gate.rs` — `Authenticate` returns "SSH session detected" while
  `TestAuthenticate` reaches the comparison loop. Alone in its own integration binary because it owns
  the process environment.
- `AuthIntent` table unit tests (charging / gates / audit source) and `decode_auth_result` sentinel
  tests.

Rewritten/deleted because the fix invalidates their premise:
- `daemon_integration.rs::exempted_authenticate_does_not_consume_rate_limit_budget` → renamed
  `test_intent_does_not_consume_rate_limit_budget`; it asserted the root-exemption premise via
  `handle_authenticate(user, false)`.
- `server_authz.rs::root_failed_attempts_do_not_charge_the_rate_limit` → **deleted**; it asserted the
  vulnerable behavior directly ("two consecutive failed ROOT attempts: neither is charged"). Replaced
  by the two charging pins above.

Both new structural pins were falsified before being kept (reintroduce the UID-0 exemption → the
regression pin fails; drop the enum variant → the matrix scan fails).

## Review finding

Addresses the High-severity finding raised independently by REV-112 and REV-BP: the N11 rate-limit
exemption keyed on caller UID disables the daemon's rate limiter for every root-run PAM stack,
including `sudo`.

### DIFF
--- a/crates/facelock-cli/src/backend.rs
+++ b/crates/facelock-cli/src/backend.rs
@@ -244,18 +244,19 @@ impl<'a> Backend<'a> {
     }
 
     /// Recognition only — backs `facelock test` (root-only, N11). NEVER the
-    /// PAM/auth path.
+    /// PAM/auth path: the daemon transport calls the root-only
+    /// `TestAuthenticate` method, not `Authenticate`.
     ///
-    /// Both transports run the same `pre_check` gates. The surviving posture
-    /// difference: the direct path skips the SSH/lid physical-presence aborts
+    /// The two transports now have the same posture: both run every
+    /// `pre_check` gate except the SSH/lid physical-presence aborts
     /// (`PreCheckContext::test` — they exist to stop an attacker's shortcuts,
-    /// not a root operator testing over SSH) and audits as `Test`, while the
-    /// daemon path enforces them and audits as `Daemon`. Neither path charges
-    /// the rate limit for a failed test: direct never records failures, and
-    /// the daemon exempts root callers.
+    /// not a root operator diagnosing over SSH), both audit as `Test`, and
+    /// neither charges the rate limit for a failed test — the direct path
+    /// because it never records failures, the daemon path because
+    /// `TestAuthenticate` is the entry point that does not.
     pub fn recognize(&self, user: &str) -> anyhow::Result<MatchResult> {
         match self.kind {
-            BackendKind::Daemon => match ipc_client::authenticate(user)? {
+            BackendKind::Daemon => match ipc_client::test_authenticate(user)? {
                 AuthOutcome::AuthResult(result) => Ok(result),
                 // `suppress_unknown` short-circuit: same "no result"
                 // shape the direct path reports.
@@ -698,7 +699,7 @@ mod tests {
             "commands/preview/wayland_preview.rs", // daemon frame loop
         ];
         let pins: &[(&str, &[&str])] = &[
-            ("authenticate", backend_only),
+            ("test_authenticate", backend_only),
             ("list_models", backend_only),
             ("remove_model", backend_only),
             ("clear_models", backend_only),
@@ -729,6 +730,28 @@ mod tests {
                     );
                 }
             }
+        }
+    }
+
+    /// The CLI must never call the daemon's `Authenticate` method. That
+    /// method is real authentication: it always charges the shared
+    /// rate-limit budget, and its only legitimate callers are the
+    /// authenticators (PAM, the polkit agent). `facelock test` is a
+    /// diagnostic and goes to the root-only `TestAuthenticate` instead —
+    /// keeping the two apart on the wire is what lets the daemon stop
+    /// guessing an attempt's purpose from the caller's UID, which is how a
+    /// failed face auth at a `sudo` prompt used to escape the limiter.
+    #[test]
+    fn the_cli_never_calls_the_real_authenticate_method() {
+        // Assembled at runtime so this test's own literal doesn't count.
+        let needle = format!("ipc_client::{}(", "authenticate");
+        for (path, content) in source_files() {
+            assert!(
+                !code_contains(&content, &needle),
+                "{}: `{needle}` is the always-charging real-auth method — \
+                 `facelock test` must call ipc_client::test_authenticate",
+                path.display()
+            );
         }
     }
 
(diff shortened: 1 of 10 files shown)


tyvsmith:

PR #127, opened 2026-08-14
chore: remove machinery that can no longer execute (review C5)

tyvsmith:
**Review finding:** REV-108 and REV-121 independently flagged that several
things kept through the waves as "cleanup candidates for later" now have
expired justifications — including a ~360-line authorization subsystem in a
security-reviewed file that no caller can reach.

Near-pure deletion: **46 insertions, 667 deletions across 24 files (net −621
lines)**. The insertions are the forced consequences — collapsed function
tails, corrected doc comments, and the two adapted test call sites. No D-Bus
wire change (the methods stay; only an internal authorization layer that can
never fire is removed), no auth-semantics change, the two protocol strings PAM
substring-matches are untouched, and `pam-facelock` and `is-enrolled` are not
touched at all.

---

### 1. The polkit frame-authorization layer in `facelock-daemon/src/server.rs`

**Unreachability proof.** `preview_detect_frame_as` opened with

```rust
authorize_method(&caller, Method::PreviewDetectFrame, None)?;
let caller_is_root = caller.uid == 0;
```

`Method::scope` routes `PreviewDetectFrame` through its catch-all arm
(`_ => Scope::Root`), so `authorize_method` calls `require_root`, which returns
`AccessDenied` for every `caller.uid != 0`. By the time control reached the
next line, `caller_is_root` was therefore **always** `true` — and the root arm
short-circuited to `FrameAuthz::Granted` without consulting the cache:

```rust
let authz = if caller_is_root {
    FrameAuthz::Granted
} else {
    /* the entire polkit path — unreachable */
};
```

Nothing else called into the cache. `run_frame_authz_check` had exactly one
call site (the `else` arm's `start_authz_check` closure) and
`watch_bus_disconnects` only evicted entries that could never be created.

**Deleted:** `PreviewAuthzCache` (with `begin_or_lookup`, `complete`, `evict`,
the TTL expiry and bounded-eviction logic), `AuthzEntry`, `FrameAuthz`,
`allow_preview_jpeg`, `PolkitAuthorizationResult`, `polkit_authority_proxy`,
`check_polkit_frame_authorization`, `run_frame_authz_check`,
`watch_bus_disconnects` plus its spawned task in `run_dbus_server`, the five
tuning constants (`PREVIEW_FRAMES_ACTION_ID`, `POLKIT_ALLOW_USER_INTERACTION`,
`AUTHZ_GRANTED_TTL`, `AUTHZ_DENIED_TTL`, `AUTHZ_CACHE_MAX_ENTRIES`,
`POLKIT_INTERACTION_TIMEOUT`), the `preview_authz` field, and the
`sender: Option<String>` / `start_authz_check: impl FnOnce(String)` parameters
that existed only to feed them. The now-unused `use std::collections::HashMap`
went with them.

**Deliberately kept:** `sanitize_preview_jpeg` and both its tests, as the
reviewers asked. It now takes `caller_is_root` directly instead of the
collapsed `allow_frames`, and its doc comment says plainly that it is a
regression hedge rather than a live gate.

**Docs corrected:** the `server.rs` module doc no longer advertises
"polkit-gated preview frames"; same for the `FacelockService` doc ("the polkit
round-trip spawn"), `commands/daemon.rs`, and two stale comments in the CLI
(`ipc_client::create_proxy`, which justified connection reuse by the polkit
cache key, and `wayland_preview`'s empty-frame branch).

**Tests deleted (9 unit tests, all covering only the deleted machinery):**
`root_gets_frames_regardless_of_polkit_state`,
`non_root_gets_frames_only_when_polkit_granted`,
`authz_cache_first_lookup_starts_a_check`,
`authz_cache_polkit_authorized_grants_until_ttl`,
`authz_cache_polkit_denied_refuses_then_allows_retry`,
`authz_cache_polkit_error_fails_closed`,
`authz_cache_eviction_kills_grant_with_connection`,
`authz_cache_drops_verdict_for_disconnected_caller`,
`authz_cache_is_bounded`.

**Tests kept:** both `server_authz.rs` integration tests still pin live
behavior and were only adapted to the shorter signature —
`every_entry_point_except_authenticate_denies_non_root` still asserts the
`PreviewDetectFrame` denial, and
`preview_detect_frame_for_root_returns_frames_and_scores` still asserts root
gets frame bytes and the unredacted score.

### 2. `dbus/org.facelock.policy`

The file declared exactly one action, `org.facelock.preview-frames`, whose only
consumer was the check deleted above. With it gone the file is empty of live
actions, so the file itself is removed.

**The `facelock-polkit` agent is a different thing and was not touched.** It
gates on `polkit.face_eligible_actions`, which defaults to
`["org.freedesktop.login1.lock-sessions"]` — it never consumed this action.
`grep -rn 'org.facelock.policy\|preview-frames\|polkit-1/actions'` over the
tree now returns nothing.

**Packaging references updated** (a stale reference here is a broken install):
`justfile` (`install-files`, `uninstall-files`, the paths listing),
`dist/PKGBUILD`, `dist/PKGBUILD-bin`, `dist/PKGBUILD-git`,
`dist/facelock.spec` (both the `install` and the `%files` entry),
`dist/debian/rules`, `dist/nix/default.nix`, `dist/nix/module.nix`,
`.github/workflows/scripts/build-deb.sh`, `.../validate-deb.sh`,
`.../validate-rpm.sh`.

In `dist/nix/module.nix` only the `environment.pathsToLink = [ "/share/polkit-1" ]`
line went (the package no longer ships anything under that path);
`security.polkit.enable = true` stays, since the package still ships the
`facelock-polkit-agent` binary that needs a polkit authority.

**Test-harness references:** `test/pkg-validate.sh` drops three assertions
("polkit action policy exists", "...declares preview-frames action", "...XML is
valid"); `test/run-integration-tests.sh` drops the `polkitd` startup/cleanup
harness and three tests ("polkit action policy installed", "PreviewDetectFrame
serves frames to polkit-authorized caller", "PreviewDetectFrame stripped again
after polkit rule removal"); `test/Containerfile` drops the `polkit` package it
installed only for those tests.

### 3. Write-only `camera_owner_uid`

Re-verified with grep: two `set_camera_owner` call sites, five
`clear_camera_owner` call sites, and **zero reads** — the only mentions of the
field were its declaration, its two constructions, and the two write helpers.
It became dead when `ReleaseCamera` went root-only and the ownership
comparison it fed was removed.

Deleting it collapses three `let result = …?; if result.is_ok() { … } result`
tails back to the plain `?` tail that the sibling methods
(`list_devices_as`, `list_models_as`, …) already use. Nothing branched on the
value, so there is no behavior change.

### 4. `TerminalNotifier` in `facelock-cli/src/notifications.rs`

Carried under `#[allow(dead_code)]` with a comment promising production wiring
"with H7/D10". Both landed and neither wanted it: the message seam that shipped
in #119 is `crate::message::Terminal`, which renders the same user-facing text
plus a machine-log side channel, and every CLI command routes through that.
`grep -rn TerminalNotifier crates/` found only the struct and its `Notifier`
impl — no constructor anywhere, including the unit tests the comment claimed
exercised it (the `notifications.rs` tests cover `body`/`icon`/`timeout_ms` and
the notifier factory, none of them this type).

### Nothing else was found live-but-suspicious

The optional scope item is a no-op: `FrameAuthz` was the only `pub` item in
`server.rs` with no external consumer, and it disappeared with target 1. Every
remaining `pub` item there is consumed — the `*_as` entry points by
`tests/server_authz.rs`, and `run` / `ServerError` / `CallerIdentity` /
`FacelockService` / the type aliases by the binary.

### Gates

All three run locally; tails in the review thread.

- `just check` — clean (test, lint, fmt-check, audit, check-pam-standalone)
- `just test-arch-pam` — 22/22
- `just test-arch-layout` — 21/21


tyvsmith:
### Gate tails (run locally on `chore/dead-code-sweep`)

**`just check`** — clean through all five sub-recipes (`test`, `lint`, `fmt-check`, `audit`, `check-pam-standalone`):

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.10s
cargo fmt --all -- --check
    Scanning Cargo.lock for vulnerabilities (477 crate dependencies)
warning: 2 allowed warnings found          # pre-existing yanked-crate advisories (core2, uds_windows)
pam-facelock crate count: 57
pam-facelock dependency guard passed
```

`cargo test --workspace`: every suite `0 failed` — 87 (daemon lib), 242 + 108 + 49 (cli), 68, 29, 29, 27, 24, 22 (server_authz), 15, 12, 11, 10, 8, 7, 4, 2, 1, 1, plus doc-tests; 10 ignored (hardware-gated).

**`just test-arch-pam` — 22/22**

```
=== PAM Container Tests ===
TEST: Module loads without crash ... PASS
TEST: Module returns gracefully when daemon not running ... PASS
TEST: Module handles missing config ... PASS
TEST: Module respects disabled config ... PASS
TEST: pam_sm_authenticate symbol exists ... PASS
TEST: pam_sm_setcred symbol exists ... PASS
TEST: facelock setup requires root ... PASS
TEST: facelock daemon requires root ... PASS
TEST: facelock auth exits 2 when no faces enrolled ... PASS
TEST: No enrolled faces: pamtester completes quickly ... PASS
TEST: PAM shows 'Identifying face...' text ... PASS
TEST: PAM respects notification mode=off ... PASS
TEST: Oneshot mode: no enrolled faces returns quickly ... PASS
TEST: Group-writable config rejected, fails closed ... PASS
TEST: World-writable config rejected, fails closed ... PASS
TEST: Config accepted again after restoring 644 ... PASS
TEST: env_clear: marker was active in the PAM process ... PASS
TEST: env_clear: LD_PRELOAD marker not loaded by oneshot child ... PASS
TEST: env_clear: SSH_CONNECTION survives to oneshot child ... PASS
TEST: env_clear: oneshot child PATH pinned to /usr/bin:/bin ... PASS

```

This tier is the real proof for the packaging half of target 2: the image is
built by `just install-files`, so it would have failed at image-build time if
any install path still referenced the deleted `dbus/org.facelock.policy`.

**`just test-arch-layout` — 21/21**

```
=== State Layout Tests ===
TEST: /var/lib/facelock is 710 root:facelock ... PASS
TEST: /var/lib/facelock/models is 755 root:root ... PASS
TEST: /var/lib/facelock/enrolled is 710 root:facelock ... PASS
TEST: /var/log/facelock is 700 root:root ... PASS
TEST: /var/log/facelock/snapshots is 700 root:root ... PASS
TEST: /var/lib/facelock is 710 root:facelock ... PASS
TEST: /var/lib/facelock/enrolled is 710 root:facelock ... PASS
TEST: /var/lib/facelock/facelock.db is 600 root:root ... PASS
TEST: no entry under the state dir is other-accessible (models/ excepted) ... PASS
TEST: the state dir itself grants 'other' nothing ... PASS
TEST: group member reads own marker through 0710 dirs ... PASS
TEST: group member cannot list the state dir ... PASS
TEST: group member cannot list enrolled/ ... PASS
TEST: group member cannot read the database ... PASS
TEST: group member can read a model file by name ... PASS
TEST: non-member cannot traverse the state dir at all ... PASS
TEST: non-member cannot list models/ either ... PASS
TEST: non-member cannot read the audit log directory ... PASS

```

Neither container test script was modified by this PR (`git diff` over
`test/run-container-tests.sh` and `test/run-layout-tests.sh` is empty), so both
counts are unchanged baselines rather than adjusted ones.

tyvsmith:

PR #122, opened 2026-08-14
refactor(status): Health facts + renderer — unknown is not false (N4, F7)

tyvsmith:
Wave 3, PR H8 — the Health domain (map §5, post-DEC-6/DEC-7; gap register N4's remnant, F7 reopened, DEC-7). `status` becomes a testable renderer over a `Health` fact model, and "unknown" stops collapsing into "false".

## Design: `Fact<T>` and `Health`

**`Fact<T>`** (`crates/facelock-cli/src/resolved.rs`) is the grown form of H2's `Resolved<T>` — adopted, not duplicated:

```rust
pub enum Fact<T> {
    Known(Resolved<T>),          // value + the Provenance every resolved fact already carries
    Unknown { why: String },     // the probe could not determine the answer
}
```

The map sketched `Unavailable { needs: Privilege }`. DEC-6 collapsed the privilege tiers — `status` is root, `is-enrolled` never touches this machinery — so what is left of N4 is *probe failure* (unreadable database, unparsed config), and the honest payload is the reason, not a privilege level. `From<Resolved<T>>` lets every H2 probe (`ModelFiles`, `CameraPresence`, `ExecutionProviderFact`) and H6's reachability discipline lift into a `Fact` unchanged.

**`Health`** (`crates/facelock-cli/src/health.rs`) is shaped from what `status` actually reports, not the map's six-field sketch:

| Field | Type | Probe |
|---|---|---|
| `config` | `ConfigHealth` (not a `Fact` — running the parse *is* the probe) | `ConfigHealth::from_load` |
| `daemon` | `Fact<DaemonPing>` | `backend::probe_daemon` (in the seam) |
| `oneshot_fallback` | `Fact<OneshotFallback>` (F7) | `probe_oneshot_fallback_at` |
| `camera` | `Fact<CameraHealth>` (presence + interrogated caps incl. `applied_quirks`) | `probe_camera` → `direct::resolve_camera_device` (H3's one implementation) |
| `models` | `Fact<ModelFiles>` (H2's type, unchanged) | `ModelFiles::probe` |
| `execution_provider` | `Fact<ExecutionProviderFact>` (H2's type) | `ExecutionProviderFact::probe` |
| `encryption` | `Fact<EncryptionHealth>` | `probe_encryption` |
| `user` + `enrolled` | `Fact<EnrollmentHealth>` (store + marker diagnostic, DEC-7) | `probe_enrolled` |
[...]

PR #128, opened 2026-08-14
refactor(cli): one fact type, no unread provenance (review C2)

tyvsmith:
## The design, and why

**Chosen: (a)** — flatten to `enum Fact<T> { Probed(T), Claimed(T), Unknown { why: String } }`, delete `Resolved<T>` and `Provenance` outright. Not (b): rendering probed-vs-claimed adds user-visible surface nobody asked for, and the reviewers' recommendation was to delete the unused thing, not to find it a job.

The one judgment call worth flagging: I kept the probed/claimed distinction as a **variant name** rather than deleting it entirely. It has exactly one reader — `Backend::select`'s `backend selected` trace, which prints the fact's `Debug`, so an operator debugging why a command went direct can see whether the bus was asked at all. As a variant name it costs nothing, it documents at each construction site whether the answer was verified, and it cannot regrow into a second level of nesting. What it no longer is: a separate type with its own field, read by nobody.

`Fact::probed`/`claimed`/`unknown` survive as constructors. `unknown` has to be a function (its variant holds an owned `String`); keeping the other two beside it means every construction site reads alike, and it is why the two money pins below compile untouched.

`DaemonReachability` and `DaemonPing` stay separate, as intended — selection's probe must never activate the daemon, `status`'s ping may. Both now carry a `Fact` instead of a `Resolved`.

## Deleted

| What | Where | Lines |
|---|---|---|
| `ResolvedConfig` + `impl` | `resolved.rs` | 18 |
| `Provenance` enum | `resolved.rs` | 14 |
| `Resolved<T>` struct | `resolved.rs` | 6 |
| `From<Resolved<T>> for Fact<T>` | `resolved.rs` | 5 |
| `"ResolvedConfig"` forbidden token | `resolved.rs` pin test | 1 |

`ResolvedConfig` had zero call sites anywhere in the tree; its doc named consumers (`daemon` startup logging, `status`) that call the individual probes directly, and #122 built `Health` rather than adopting it. It escaped the dead-code lint only by being `pub` in a `pub mod`. The remaining forbidden token `resolved::` still covers it in the `is-enrolled` isolation pin.

Falling out of the collapse: the three probes in `resolved` are infallible, so they return their value plainly instead of a wrapper that could only ever say "known". Four call sites in `daemon`/`enroll`/`test_cmd` lose a `.value` hop.

Net across the branch: **275 insertions, 259 deletions** — but that includes ~90 lines of new tests and docs; the type machinery itself is a net deletion.

## Bug 1 — a probed absence was tagged as a claim

`health.rs`, `probe_camera` (was ~:255-259). With no `device.path`, `status` runs auto-detection; if detection came back empty, the fact was tagged `Claimed` — "taken from the config without verification" — but the config says nothing to take, and the emptiness is precisely what the probe observed. The other three paths through the same function were `Probed`, so this was drift, not a distinction. Nothing read the tag, so nothing caught it.

```rust
// before
None => Resolved {
    value: CameraHealth::AutoDetect { resolved: None },
    provenance: crate::resolved::Provenance::Claimed,
},
some => Resolved { value: CameraHealth::AutoDetect { resolved: some }, .. Probed },

// after — every arm is Probed, which collapses the inner match entirely
CameraPresence::AutoDetect => Fact::probed(CameraHealth::AutoDetect {
    resolved: interrogate(),
}),
```

The rule moved behind `camera_fact(presence, interrogate)` with the device interrogation injected — the split `backend::classify` already uses — so it is testable without hardware. New tests: `auto_detect_finding_nothing_is_a_probed_absence`, `an_absent_configured_node_is_never_interrogated`.

## Bug 2 — an unknown fact could collapse to a confident `false`

`health.rs`, `probe_oneshot_fallback_at` (was ~:99-104). `Fact::known()` answers `None` for an unknown fact, so the derivation below hands `models_present: false` to the report whenever the model fact is *undetermined* — rendering "models: missing / fallback not usable", a guessed false answer from a fact nobody established. That is the N4 rule this layer exists to enforce, broken at the layer's own call site.

```rust
// before — bool argument, supplied by folding a Fact through a default
let models: Fact<ModelFiles> = ModelFiles::probe(config).into();
let fallback = probe_oneshot_fallback_at(
    Path::new(AUTH_BIN), Path::new(&config.storage.db_path),
    models.known().is_some_and(|m| m.all_present()),
);

// after — derive from the value, then lift; the signature takes &ModelFiles
let models = ModelFiles::probe(config);
let fallback = probe_oneshot_fallback_at(
    Path::new(AUTH_BIN), Path::new(&config.storage.db_path), &models,
);
// ... models: Fact::probed(models)
```

The collapse is now unrepresentable rather than merely unreachable: `models.known().is_some_and(..)` is a `bool` and no longer type-checks against `&ModelFiles`, and at the point `all_present()` is called there is no `Fact` in scope at all.

**On the test.** No value-level test can distinguish the two shapes — that is exactly what made the bug latent, since `ModelFiles::probe` cannot fail today. So `probe_derives_the_fallback_from_the_model_value_not_a_fact` is a source-level pin over `Health::probe`'s body (the idiom `resolved.rs` already uses twice), and it does fail on the old code: the old body contains `known()`, the new one does not. The signature carries the guarantee; the pin guards the call site. `fallback_is_usable_only_with_binary_models_and_database` was also widened to assert `models_present` directly.

## Rendered output

**Unchanged.** `healthy_report_renders_byte_exact` passes untouched, as does every other renderer test. Bug 1 changes only a tag that was never rendered; bug 2 was latent. No golden update was needed.

Both money pins pass **unmodified**: `commands::status::tests::unreachable_daemon_and_unknown_store_never_render_as_not_enrolled` and the out-of-process `tests/health_render.rs`.

## Gates

- `just check` — 245 cli lib tests + all workspace tiers, clippy `-D warnings`, fmt, audit, pam dependency guard: clean

No auth-semantics change, no D-Bus wire change, `is-enrolled` untouched.

Review finding: cross-cutting **C2** (four reviewers converged: more fact/provenance structure than consumers), plus REV-116 and REV-122 (findings 1 and 3).

### DIFF
--- a/crates/facelock-cli/src/backend.rs
+++ b/crates/facelock-cli/src/backend.rs
@@ -40,7 +40,7 @@ use facelock_daemon::auth::AuthOutcome;
 use facelock_store::StoreError;
 
 use crate::message::{Terminal, UserMessage};
-use crate::resolved::{Provenance, Resolved};
+use crate::resolved::Fact;
 use crate::{direct, ipc_client};
 
 /// Which implementation answers this invocation's questions — and, for the
@@ -64,9 +64,9 @@ impl BackendKind {
     }
 }
 
-/// What the one probe observed, recorded beside the resolved-config facts
-/// (D7) with the same provenance discipline: `NotProbed` is a distinct value,
-/// never a guessed `Unreachable`. `status` reports the same discipline
+/// What the one probe observed, recorded as a [`Fact`] beside the config
+/// facts in `resolved` (D7) and with the same discipline: `NotProbed` is a
+/// distinct value, never a guessed `Unreachable`. `status` reports it
 /// through [`DaemonPing`], which additionally carries the failure detail a
 /// report wants and a human does not need at selection time.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -339,7 +339,7 @@ pub enum DaemonPing {
 
 /// `status`'s daemon probe. Lives in the seam so `ipc_client::ping` stays
 /// single-sited (D1) — the health model consumes the fact, not the transport.
-pub fn probe_daemon(mode: &DaemonMode) -> Resolved<DaemonPing> {
+pub fn probe_daemon(mode: &DaemonMode) -> Fact<DaemonPing> {
     probe_daemon_with(mode, ipc_client::ping)
 }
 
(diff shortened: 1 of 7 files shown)


tyvsmith:

PR #129, opened 2026-08-14
refactor(auth): carry rejection class as a type, not as prose (review C4)

tyvsmith:
Two defects that share one disease: a value that should have been a type was
carried as English prose, and four independent matchers read it.

## Defect 1 — the rejection class was free text

A recoverable rejection has four consequences. All four were derived by
substring-matching the same sentence:

| Consequence | Before | After |
|---|---|---|
| Audit `result` label | `message.contains("rate limited")` — `facelock-daemon/src/auth.rs:216` | `kind.audit_result()` — `auth.rs:341` |
| Oneshot exit code | `message.contains("all frames dark")` — `facelock-cli/src/commands/auth.rs:218` | `oneshot_exit_code(kind)` — `commands/auth.rs:226,286` |
| CLI wire decode | `AuthOutcome::Error { message: label }`, class lost — `facelock-cli/src/ipc_client.rs:241` | `ErrorKind::classify(&label)` — `ipc_client.rs:249` |
| PAM return code | `message.contains("rate limited")` / `"IR camera required"` — `pam-facelock/src/lib.rs:865-873` | **unchanged, deliberately** |

Rewording a message changed PAM policy and mislabelled the audit trail,
silently, with no test spanning the four. `AuthOutcome`'s own doc comment
admitted the strings were "protocol-bearing" — a comment doing a type's job.

`AuthOutcome::Error` now carries `{ kind: ErrorKind, message: String }`, and
`ErrorKind::render` is the only place any of these sentences is written.

### The class set, and who produces each

Derived from the actual producers, not invented:

| Class | Producer |
|---|---|
| `Disabled` | `pre_check_with_context`, `security.disabled` |
| `SshSession` | `pre_check_with_context`, `abort_if_ssh` |
| `LidClosed` | `pre_check_with_context`, `abort_if_lid_closed` |
| `Storage` | `pre_check_with_context`, `store.has_models` failure |
| `RateLimited` | `pre_check_with_context`, budget exhausted |
| `RateLimitCheckFailed` | `pre_check_with_context`, the limiter itself errored |
| `IrRequired` | `pre_check_with_context`, `require_ir` vs non-IR device |
| `AllFramesDark` | `authenticate_inner`, every frame below the dark threshold |
| `Internal` | `ErrorKind::classify` only — a message from a daemon of another version |

Two things the brief listed are deliberately absent. `NotEnrolled` is not an
`Error` at all: it is `AuthResult { matched: false }` or `Suppressed`, so
giving it a class would have invented a producer. And `RateLimitCheckFailed`
is split out from `RateLimited` because a limiter that failed to answer is a
storage fault, not a lockout — the old substring matcher got that right only
by accident of wording (`"rate limit check failed: …"` happens to contain
neither `"rate limited"` nor `"rate_limit"`). It is now right by construction.

### The frozen strings are byte-identical

PAM's matcher is untouched, so the daemon must keep producing the exact bytes
it matches. Evidence:

- `frozen_protocol_strings_render_byte_exactly` (`facelock-daemon/src/auth.rs`)
  asserts `render` returns `"rate limited"` and the full `"IR camera required
  for authentication. Set security.require_ir = false to override (NOT
  RECOMMENDED)."` verbatim.
- `only_the_two_intended_classes_trip_pams_matchers` replicates PAM's two
  predicates and asserts the converse — that **no other** class trips them.
  This is the direction that was previously invisible; `pam-facelock` cannot
  link `facelock-daemon` (dependency ceiling: libc/toml/serde/zbus), so
  nothing else can catch the two classifications drifting apart.
- The pre-existing wire pins in `tests/server_authz.rs`
  (`rate_limited_flows_in_band_with_the_exact_protocol_string`,
  `require_ir_flows_in_band_with_the_exact_protocol_string`) still pass
  unmodified — I did not touch them, and they assert `result.label ==
  "rate limited"` at the D-Bus boundary.
- Mutation check: rewording `RateLimited` to `"too many attempts"` fails
  `every_rejection_class_pins_its_message_audit_label_and_exit_code` and
  `recoverable_error_sentinel_decodes_with_the_message_intact`.

`ErrorKind::classify` is the inverse of `render`, pinned round-trip by
`classify_inverts_render`. It is now the only text match on the daemon/CLI
side of the wire.

## Defect 2 — the redaction sentinel aliased PAM's "was a face seen?" signal

PAM chose `PAM_IGNORE` vs `PAM_AUTH_ERR` on a non-match by testing
`similarity == 0.0`, reading a zero score as "the camera saw nobody". But
#108's N12 redaction zeroes the score for **every** non-root caller, so for a
user-run locker (hyprlock) a genuine face-seen non-match was byte-identical to
an empty frame and wrongly abstained. #108 spotted this and deferred it to
#109; #109 never carried it. It cannot grant access — a match drives success
independently and both branches degrade — so this is a correctness and
telemetry defect, not a bypass.

### Sentinel design

`MatchResult::face_detected` records whether the detector ever found a face,
set in `authenticate_inner` right after the `faces.is_empty()` check —
detection, not recognition, so a face rejected by the IR texture gate or
dropped by device coupling still counts as seen. It reaches the wire as
`model_id == -4` via `wire_model_id`, using the sentinel space the protocol
already documents rather than a struct field, so **the D-Bus signature is
unchanged**.

What `model_id` carried before, verified rather than assumed: every non-match
sent `-1` unconditionally (`result.model_id.map(…).unwrap_or(-1)`), and the
face-seen information existed only in `similarity`. `-1`, `-2` and `-3` keep
their exact meanings; `-4` is purely additive. Documented in
`docs/contracts.md`.

Both skew directions are byte-compatible, which is why this shape was chosen
over redefining `-1`:

- **new daemon, old PAM**: `-4` falls through that module's sentinel `match`
  to the same arm as `-1` → today's behavior exactly.
- **old daemon, new PAM**: `-1` carries no face-seen signal and the module
  falls back to the score test it used before → today's behavior exactly.

Only new/new changes, and only to become correct. The realistic window — a
package upgrade whose daemon has not restarted — is the second case.

`face_detected` is a *detector* signal, not a matcher signal: it says a face
was present, never how close it came to an enrolled template. It is therefore
not a hill-climbing oracle, which is why it is not redacted while `similarity`
is.

### The pam-facelock diff, line by line

Four hunks, no new dependencies (the crate-count guard still reports 57):

1. `enum AuthResponse` — adds `NoMatchFaceSeen` (no fields). Needed because
   the decoded reply must be able to represent the new sentinel; adding a
   field to `NoMatch` instead would have touched every existing arm.
2. `parse_auth_reply` — adds `-4 => AuthResponse::NoMatchFaceSeen`. One arm,
   above the existing `_` fallback, so every other sentinel decodes unchanged.
3. `pam_code_for_no_match(face_seen, similarity)` — new. The decision that was
   inline in `identify()`. Extracted, not rewritten: the `similarity == 0.0`
   branch is verbatim, now reachable only when the reply carries no face-seen
   signal. Extracted so the fix is testable at all — `identify()` needs a live
   `pamh`. It is a sibling of the existing `pam_code_for_daemon_error`, which
   exists for exactly this reason, so this follows the file's own pattern.
4. `identify()` — the two `NoMatch` arms now call it. Behavior for
   `NoMatch { similarity }` is unchanged.

Every other PAM behavior is identical: `pam_code_for_daemon_error` and its
frozen matcher, the oneshot fallback, peer verification, timeouts, logging.

## Tests

- `every_rejection_class_pins_its_message_audit_label_and_exit_code`
  (`facelock-cli/src/commands/auth.rs`) — **the coupling test**. One table:
  every class × (exact wire message, audit label, oneshot exit code), plus an
  assertion that every `ErrorKind::ALL` variant has a row. `render`,
  `audit_result` and `oneshot_exit_code` are all exhaustive matches, so a new
  class cannot silently inherit a neighbour's policy.
- `frozen_protocol_strings_render_byte_exactly` and
  `only_the_two_intended_classes_trip_pams_matchers` — the frozen-string pins
  in both directions.
- `classify_inverts_render` — the wire decode is the exact inverse of the
  renderer.
- `a_failed_rate_limit_check_is_not_a_rate_limit_rejection` — pins the split.
- `a_redacted_caller_can_tell_a_seen_face_from_an_empty_frame`
  (`tests/server_authz.rs`) — defect 2 through the real service and the real
  redaction: a non-root caller's face-seen non-match returns `(-4, 0.0)`, an
  empty-frame non-match returns `(-1, 0.0)`, and the two replies differ.
  **Covers**: the daemon's reply as an unprivileged caller actually sees it,
  including redaction. **Does not cover**: PAM itself, a separate binary that
  cannot be linked into this test.
- `a_redacted_face_seen_no_match_is_not_read_as_an_empty_frame`
  (`pam-facelock`) — the other half: both replies carry `similarity == 0.0`,
  as a redacted caller's always do, and only the sentinel separates them.
  Plus `no_match_without_a_face_seen_signal_falls_back_to_the_score` (the
  old-daemon window) and `no_match_never_maps_to_success`.
- `a_match_still_reports_its_model_id_and_score_to_root` — the `-4` sentinel
  does not leak into a matched reply.

Both fixes were mutation-checked: removing the `-4` arm makes the daemon test
fail with `model_id: -1, similarity: 0.0` — literally the aliased reply.

## Recommendation NOT implemented here

The oneshot path collapses **every** pre-flight rejection to exit 2 →
`PAM_IGNORE`, while the daemon path distinguishes rate-limited
(`PAM_AUTH_ERR`) and suppressed (`PAM_AUTHINFO_UNAVAIL`). So whenever the
daemon is unavailable and PAM falls back to the oneshot, rate-limiting's PAM
consequence silently softens — under `required`/`requisite` stacks that is a
real semantic difference. This PR does not change it: it changes PAM
semantics and deserves its own reviewed change.

`ErrorKind` makes it a one-table edit. `oneshot_exit_code` is already the
single place both call sites consult (I routed the pre-flight short-circuit
through it here, which changes nothing today because no pre-flight gate can
produce `AllFramesDark`). Codes I would reserve:

| Exit | Meaning | Classes | PAM maps to |
|---|---|---|---|
| 0 | matched | — | `PAM_SUCCESS` |
| 1 | no match / no usable image | `AllFramesDark` | `PAM_AUTH_ERR` |
| 2 | error, no opinion | `Disabled`, `SshSession`, `LidClosed`, `IrRequired`, `Storage`, `RateLimitCheckFailed`, `Internal` | `PAM_IGNORE` |
| 3 | **rate limited** | `RateLimited` | `PAM_AUTH_ERR`, no retry |
| 4 | **suppressed** | `AuthOutcome::Suppressed` | `PAM_AUTHINFO_UNAVAIL` |

Size: ~15 lines in `oneshot_exit_code` plus the `Suppressed` arm, ~20 in
`run_oneshot_auth`'s exit-code mapping in `pam-facelock`, one row per code in
the `docs/contracts.md` PAM table, and a container test per new code. The risk
is entirely in the PAM half — 3 and 4 must not reach a module that reads them
as "no match".

## Constraints honored

No auth-decision change. No D-Bus method signature change. `is-enrolled`
untouched. `AuditSource` semantics untouched. `just check` (clippy
`-D warnings` on facelock-cli's lib target) passes.

## Gates

- `just check` — pass (clippy `-D warnings`, fmt, audit: 2 pre-existing
  allowed yanked-crate warnings, pam dependency guard: **57 crates**,
  unchanged)

Review finding: REV-109 / REV-108 / REV-BP (C4).

### DIFF
--- a/crates/facelock-cli/src/backend.rs
+++ b/crates/facelock-cli/src/backend.rs
@@ -265,6 +265,7 @@ impl<'a> Backend<'a> {
                     model_id: None,
                     label: None,
                     similarity: 0.0,
+                    face_detected: false,
                     failure_reason: None,
                 }),
                 AuthOutcome::Error { message, .. } => bail!("daemon error: {message}"),
--- a/crates/facelock-cli/src/backend.rs
+++ b/crates/facelock-cli/src/backend.rs
@@ -267,7 +267,7 @@ impl<'a> Backend<'a> {
                     similarity: 0.0,
                     failure_reason: None,
                 }),
-                AuthOutcome::Error { message } => bail!("daemon error: {message}"),
+                AuthOutcome::Error { message, .. } => bail!("daemon error: {message}"),
             },
             _ => direct::authenticate(self.config, user),
         }
--- a/crates/facelock-cli/src/commands/auth.rs
+++ b/crates/facelock-cli/src/commands/auth.rs
@@ -11,7 +11,7 @@ use facelock_core::types::CameraCaps;
 use facelock_core::types::MatchResult;
 use facelock_daemon::audit::{self, AuditEntry, AuditSource};
 use facelock_daemon::auth;
-use facelock_daemon::auth::AuthOutcome;
+use facelock_daemon::auth::{AuthOutcome, ErrorKind};
 use facelock_daemon::rate_limit::RateLimiter;
 use facelock_face::FaceEngine;
 use facelock_store::FaceStore;
@@ -113,8 +113,16 @@ pub fn run(user: String, config_path: Option<String>) -> i32 {
         // password rather than registering a failed match (exit 1). The
         // suppressed / not-enrolled / rate-limited distinction is carried by
         // the audit record and the tracing output, not the exit code.
+        //
+        // `oneshot_exit_code` is the single table, so the recommendation to
+        // stop collapsing these classes has one place to land. It cannot
+        // change anything today: no pre-flight gate produces the one class
+        // that maps to 1 (the camera is not open yet).
         debug!(?resp, "pre-check short-circuit");
-        return 2;
+        return match resp {
+            AuthOutcome::Error { kind, .. } => oneshot_exit_code(kind),
+            _ => 2,
+        };
     }
 
     // Quirk override takes precedence over the config's warmup value.
@@ -215,19 +223,25 @@ pub fn run(user: String, config_path: Option<String>) -> i32 {
             );
             1
         }
-        AuthOutcome::Error { message } if message.contains("all frames dark") => {
+        AuthOutcome::Error {
+            kind: ErrorKind::AllFramesDark,
+            ..
+        } => {
+            // `authenticate_inner` already audited this one, so unlike the
+            // arm below there is nothing to write here.
             info!(user = %user, "all frames dark");
-            1
+            oneshot_exit_code(ErrorKind::AllFramesDark)
         }
-        AuthOutcome::Error { message } => {
-            // Errors from authenticate() that aren't "all frames dark" are storage errors
-            // which happen before the auth loop — audit those here.
+        AuthOutcome::Error { kind, message } => {
+            // Errors from authenticate() other than all-frames-dark are
+            // storage errors which happen before the auth loop — audit those
+            // here, under the class's own label.
             audit::write_audit_entry(
                 &config.audit,
                 &AuditEntry {
                     timestamp: audit::now_iso8601(),
                     user: user.clone(),
-                    result: "error".into(),
+                    result: kind.audit_result().into(),
                     source: Some(AuditSource::Oneshot),
                     similarity: None,
                     frame_count: None,
@@ -238,7 +252,7 @@ pub fn run(user: String, config_path: Option<String>) -> i32 {
                 },
             );
             error!(user = %user, "auth error: {message}");
-            2
+            oneshot_exit_code(kind)
         }
         _ => {
             error!("unexpected response");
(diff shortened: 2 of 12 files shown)


tyvsmith:

PR #130, opened 2026-08-14
refactor(i18n): split the message vocabulary and validate translations (review C3)

tyvsmith:
Review finding: REV-119 (message seam durability), REV-BP (i18n validation gap).

`crates/facelock-cli/src/message.rs` was 1,889 lines holding one flat
`UserMessage` enum of **200 variants**, maintained as three parallel lists in
one file — the enum, the `localized()` match, and the sample list. It is the
file every feature has to touch, PR #122 had to invent a `Status*` prefix to
namespace its 55 additions, and the ~385 raw print sites still waiting to be
converted would have roughly doubled it. Alongside that, `po/facelock.pot`
carried 201 msgids and **zero** format flags, so `msgfmt --check` validated
nothing about the `{brace}` placeholders the seam fills.

## The taxonomy

Nine per-domain types behind one `Message` trait carrying the two renderings
(`localized()` / `machine()`). The sinks (`Terminal`, `fail`, `explain`), the
gettext plumbing (`init`, `translate`, `fill`) and the trait stay shared in
`message/mod.rs`.

| Module | Variants | What it says |
|---|---:|---|
| `access` | 10 | config load, privilege, which backend answered |
| `face` | 32 | `enroll` / `test` / `remove` / `clear` |
| `status` | 55 | the `facelock status` report's prose |
| `notify` | 4 | notification bodies (rendered `NotifyEvent`) |
| `setup` | 47 | wizard spine, step failures, closing summary |
| `device` | 12 | camera and inference-device selection |
| `download` | 9 | model download and verification |
| `system` | 11 | daemon unit and `facelock` group setup |
| `pam` | 20 | `/etc/pam.d/*` configuration |

The boundaries are the section banners the file had already grown, not a
taxonomy imposed on it — the only judgement calls were splitting the
"enrollment / test / systemd steps" banner (the four wizard prompts to
`setup`, the five systemd events to `system`) and giving group membership and
model download their own modules rather than leaving them inside a 106-variant
`setup`. Largest domain file is 336 lines, down from 1,889; `mod.rs` is 613 — the FFI
wrapper, the trait, the sinks and the shared tests.

**Variant names are unchanged**, so `StatusMessage::StatusHeader` still
stutters. That was deliberate: `machine()` is a derived `Debug` rendering, so
renaming variants would change the tracing event stream, which is a behaviour
change this PR is not chartered to make. Dropping the now-redundant prefixes
is a clean follow-up once someone decides that stream may move.

## Byte-identity evidence

A temporary test dumped `machine()` + `localized()` for every variant before
and after the split. All **200 records identical**, as multisets and as
`machine()` line sets:

```
records before=200 after=200
identical as multisets: True
machine() lines identical: True 200
```

`just check` (test + lint + fmt-check + audit + pam-standalone) passes, which
includes the status renderer's golden byte-exact test.

## .pot msgid-set evidence

`just pot` now globs the domain modules (`ls .../message/*.rs | LC_ALL=C sort`,
xgettext takes multiple inputs). After the split, sorted msgid sets before vs.
after:

```
$ diff msgids-before.txt msgids-after-split.txt
$ wc -l msgids-after-split.txt
201    # 200 msgids + the header
```

Empty diff. Only the `#:` source references move. After the whole branch,
exactly one msgid differs — the fix-3b change, isolated:

```
163c163
< msgid "Root required. Re-run with sudo? [Y/n] "
---
> msgid "Root required. Re-run with sudo?"
```

## Placeholder validation

62 templates carry placeholders; all 62 are now flagged `python-brace-format`,
whose syntax is exactly the seam's (`{lower_snake}`, see `fill`).

**The flag in the brief does not work.** `--flag=translate:1:python-brace-format`
is accepted and silently dropped: xgettext honours a format flag only for the
format types its *scanner language* knows, and `--language=C` knows only
`c-format`. Verified against gettext 1.0 on the same input — `--language=Python`
emits the flag, `--language=C` emits nothing. Scanning Rust as Python is not an
option (`///` doc-comment apostrophes would open Python string literals), so
the flags are applied after extraction by matching the seam's one placeholder
syntax, and then `just pot` makes gettext confirm its own work:

```
msgen po/facelock.pot | msgfmt --check-format -o /dev/null -
```

Every flagged msgid must itself parse as a brace-format string, so a msgid with
an unbalanced brace in prose fails at extraction rather than in a translator's
catalog.

### msgfmt rejects a bad placeholder

Hand-broken `po/xx/facelock.po` covering all three failure modes — a typo, a
dropped placeholder, an invented one:

```
$ just mo
po/xx/facelock.po:22: a format specification for argument 'path' doesn't exist in 'msgstr'
po/xx/facelock.po:27: a format specification for argument 'error' doesn't exist in 'msgstr'
po/xx/facelock.po:341: a format specification for argument 'host', as in 'msgstr', doesn't exist in 'msgid'
msgfmt: found 3 fatal errors
error: Recipe `mo` failed with exit code 1
```

The same catalog with the `#, python-brace-format` lines stripped — i.e. the
behaviour before this PR — compiles with exit 0 and no diagnostics.

### The pin list can no longer fall behind

`sample_messages()` was a hand-written `Vec` that nothing forced to stay
complete. Each domain now implements `Samples` as a wildcard-free `match`
walking variant to variant, so the sample *is* the arm and there is no second
list. Adding a variant without one does not compile:

```
error[E0004]: non-exhaustive patterns: `&download::DownloadMessage::ProofVariantWithNoSample` not covered
   --> crates/facelock-cli/src/message/download.rs:96:20
```

## The four smaller fixes

**a. `setlocale` is MT-Unsafe** — it was declared inside the `safe fn` block of
the gettext `extern`, which is a false claim: glibc marks it MT-Unsafe
(const:locale) while `dgettext`/`bindtextdomain` are MT-safe. Now a plain
`unsafe extern` fn (`message/mod.rs:130`) with SAFETY notes at both call sites:
`init` (`mod.rs:144`) can appeal to running once from `main` while the process
is single-threaded; the `.mo` fixture test (`mod.rs:514`) cannot, so it takes
the `LOCALE_MUTATION` mutex (`mod.rs:564`) that serializes locale-mutating
tests against each other. Practical crash risk was low; the false safety claim was the defect.

**b. Answer hint out of the msgid** — the sudo re-exec prompt localized
`"Root required. Re-run with sudo? [Y/n] "` while the parser stayed English and
defaulted to yes. The hint is appended by the sink now
(`Terminal::confirm_default_yes`, `mod.rs:251`, composing through `prompt_line`
at `mod.rs:260`), exactly as `confirm()` already did, and the rule lives at the
parsing site (`mod.rs:264-274`) and at the call site (`ipc_client.rs:33-38`),
not only in a review thread. Rendered bytes are
unchanged, pinned by `answer_hints_are_appended_not_translated`.

**c. `machine()` is not a serialization format** — the module doc
pre-authorized its derived-`Debug` output for `audit.jsonl`, where a field
rename would change the record silently. Amended at `mod.rs:10-17` and on the
trait method (`mod.rs:203-210`): persisted records go through the audit crate's
serde path; `machine()` is the tracing side channel only.

**d. Cycle between the sink and `ipc_client`** — `Terminal::confirm` called
into `ipc_client` for stdin while `ipc_client` imports the message types. The
read/parse moved into the sink next to the prompt that produced it
(`read_answer`, `mod.rs:274`); `ipc_client::confirm` is gone. Related: three
`.context(...)` sites attached `.localized()` directly, so those messages
reached stderr but never the machine event stream `Terminal`/`fail` always
emit — `ipc_client.rs:133`, `:200`, `:202` now go through `explain`
(`mod.rs:289`), which does both.

## Gates

```
$ just check
pam-facelock crate count: 57
pam-facelock dependency guard passed

$ just test          # workspace, no failures anywhere
test result: ok. 247 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ just test-arch-pam

$ just test-arch-layout
```

Out of scope and untouched: `pam-facelock`'s own gettext (its 2-entry catalog
and dependency ceiling), `is-enrolled`, D-Bus wire strings, auth semantics.
`po/pam_facelock.pot` is deliberately left alone — `just pot` regenerates it,
but its line references were already stale on the base branch and updating
them here would be unrelated churn.


tyvsmith:

Issue #99, opened 2026-08-08
Quirk name_pattern uses substring matching: 'Sirius Camera' matches '(?i)ir.*camera' and gets force_ir


tyvsmith:
Found during the #89 review series (IR hardware research).

## Problem

The H1 fix made *heuristic* IR classification token-based (`has_ir_name_token` — the "ir" in "Sirius" no longer counts). But quirk matching (`name_matches` in `crates/facelock-camera/src/quirks.rs`) is substring/regex-based, and the shipped default `config/quirks.d/00-defaults.toml` carries `name_pattern = "(?i)ir.*camera"` with `force_ir = true`.

`"Sirius Camera"` matches that pattern (the `ir` inside "Sirius", then "camera") → authoritative `IrSource::Quirk` → bypasses the token rule entirely. The corpus test asserting "Sirius" is not-IR runs **without** the quirks DB, so it can't catch this.

## Suggested fix

Either apply token semantics inside pattern evaluation for the ir/infrared atoms, or tighten the shipped pattern (e.g. `(?i)\bir\b.*camera` / explicit alternation of known IR camera names) — and add a corpus test that loads the real shipped defaults file so regressions in the pattern itself are caught.

PR #131, opened 2026-08-14
polish(camera,docs): quirk precedence, matchable patterns, honest caps (review polish P1)

tyvsmith:
Polish pass over the camera subsystem and the security docs, from the principal-architect review of #107–#122 plus two findings from the fix waves. Ten items, each re-verified against the current tree before changing anything.

**Review finding:** REV-110 (items 1, 2), REV-117 (item 3), plus two fix-wave findings (items 9, 10). Items 4–8 are the lower-severity cleanups from the same review.

---

### 1. Quirk precedence was backwards (MEDIUM, pre-existing)

`QuirksDb::load` appends `/usr/share/facelock/quirks.d` then `/etc/facelock/quirks.d`, and both the `load()` doc and the shipped TOML comments direct operators to `/etc` to correct a bad shipped entry. But `find_match_with_kind` returned the FIRST hit in vector order in both passes, so the shipped entry always shadowed the override. This matters more since #110 made `force_ir = false` authoritative unconditionally: the entry an operator writes precisely to disable a bad shipped `force_ir` was the one most reliably ignored, silently.

Matching is now last-wins within each pass, so load order is the override mechanism the docs already promise. A USB-ID match still beats a name match — that is specificity, independent of load order.

**Pins it:** `quirks::tests::etc_override_beats_shipped_quirk_for_same_device` — writes a shipped entry and an `/etc` entry for the same device into two real directories, loads them through `load_dir` in `load()`'s order, and asserts the `/etc` entry wins in *both* the name pass and the USB-ID pass.

### 2. Shipped patterns could not match the hardware their notes name (MEDIUM)

After #99's anchoring, `(?i)ir camera.*` requires the card string to START with "ir camera" — but real ThinkPad IR nodes enumerate as `Integrated IR Camera...`. The Surface/HP/Dell entries are unmatchable the same way against names that lead with a vendor word. The failure is silent: the device still works, it just loses `warmup_frames`.

The matcher has no alternation, so each affected token now ships a second `.* `-prefixed entry alongside the start-anchored one. That prefix is space-bounded, so the decoy corpus is still rejected — a bare leading `.*` would not be.

**Pins it:** the table test is now keyed on the exact pattern string rather than a substring of it (the old keying let several patterns collapse onto one case, so a pattern could be added with no name ever asserted against it), and a new `shipped_quirk_name_patterns_match_real_hardware_names` asserts real hardware-reported card strings — including `Integrated IR Camera` and the 31-char-truncated `Integrated IR Camera: Integrate` — are each matched by at least one shipped pattern, with the decoys still rejected by all of them.

### 3. Stale caps on daemon camera reopen (LOW)

The camera factory's `Some(resolved)` arm cloned the startup interrogation for every reopen, so a device swapped in at the same `/dev/videoN` after startup inherited the startup device's `is_ir` and fingerprint — caps it never demonstrated. The `None` arm already resolved fresh per open and was the honest one; both now do, so the two arms collapse into one. Reopens happen at idle-timeout frequency, not per frame.

**Pins it:** no unit test — the factory closure is built inside `build_handler`, which needs a real device, a loadable ONNX engine and a writable store. The behaviour is now structural (there is no longer a code path that replays a stale interrogation) rather than conditional.

### 4. Two `CameraCaps` fields with no production consumer (LOW)

Both dropped. `has_emitter` was asserted only by its own tests; the emitter path reads `EmitterXuInfo::from_quirk` off the matched quirk directly, and the name overstated a value that was quirk-*declared*, not probed. Renaming it to `emitter_declared` would have kept a field nothing reads, so deletion won. `formats` duplicated `DeviceInfo.formats`, which every `ResolvedCamera` holder already has, and was re-cloned on every handler build.

Every construction site outside `interrogate` uses `..Default::default()`, so no caller needed changing.

**Pins it:** the two caps tests that asserted on the removed fields now assert the surviving source of truth — `resolved.info.formats` for the format set, `EmitterXuInfo::from_quirk(resolved.quirk)` for the emitter declaration.

### 5. The unqueryable-device caps state was built inline twice (LOW)

`daemon.rs` and `auth.rs` both constructed `CameraCaps { fingerprint: device_fingerprint(..), ..Default::default() }` with a comment explaining the fail-closed semantics. Now `CameraCaps::unqueryable(fingerprint)`, documented once.

**Pins it:** `caps::tests::unqueryable_caps_fail_closed_but_keep_identity` — asserts `is_ir` is false (so `require_ir` rejects rather than admits), no quirks are recorded as applied, and the fingerprint survives, since sysfs identity is readable from the path even when the V4L2 query fails and dropping it would turn every unqueryable camera into a device-coupling mismatch.

### 6. Three public spellings of the darkness check (LOW)

`Camera::is_dark` delegated to the free `is_dark`, which hardcoded the same `0.6`/`10` that `device.dark_threshold` and `device.dark_pixel_value` default to. Production reads neither — auth and enroll both call `is_dark_with_config` with the operator's configured values. A second set of constants for the same decision drifts silently the day the config defaults change, so both default-threshold spellings are gone and `is_dark_with_config` is the only one left. `facelock-core`'s trait doc, which cited the free function as the reason darkness is not a trait method, points at it now.

**Pins it:** the two deleted tests were exact duplicates of the `*_with_config` tests beside them; coverage is unchanged.

**Not done — reported instead:** `MockCamera::is_dark` (`facelock-test-support`) still uses divergent thresholds (`ratio > 0.4`, and empty-frame → dark rather than not-dark). That crate is another agent's scope in this wave, so it is untouched here.

### 7. Double enumeration / ignored parameter (LOW)

`ResolvedCamera::resolve` took a `&QuirksDb` but `auto_detect_device` loaded its own, so the quirks were read twice and the detection pick could in principle use a different quirk set than the recorded classification. Added `auto_detect_device_with(&QuirksDb)`; `auto_detect_device` keeps its signature and loads one.

While adopting it I found the same double-load at two more call sites I own: `daemon.rs` loads a DB and then calls the no-arg `auto_detect_device`, and `auth.rs` auto-detected *before* loading its DB. Both now lend the DB they already have (`auth.rs`'s load moved above the detection).

**Pins it:** no direct test — `auto_detect_device*` enumerates `/dev/video*`, so it cannot be exercised hermetically. The parameter is now actually threaded rather than accepted and ignored.

**Not done — reported instead:** `direct.rs:67` has the same pattern; it is another agent's file this wave.

### 8. Speculative public API (LOW)

`IrEvidence` and `derive_ir_from_evidence` were `pub`, never re-exported, and had no caller outside `device.rs` — scaffolding for a `CameraCaps.is_ir` that ended up adopting `is_ir_camera_resolved`. A one-bool struct plus a query and a derive function fold into the single private predicate the two call sites wanted; the security rationale in the doc comment is preserved on it.

### 9. Security doc described the pre-N13 authorization matrix (MEDIUM, docs)

`docs/security.md` §4A still listed `Authenticate`/`ListModels`/`PreviewDetectFrame` as "root or the matching Unix user" and `ReleaseCamera` as scoped to the owner of the active preview camera session — a concept deleted outright in #127. It contradicted the corrected A5 passage ~46 lines below it in the same file, and `docs/contracts.md`, which is right.

Corrected against `Method::scope` in `facelock-daemon/src/server.rs`: `Authenticate` is the only user-scoped method, everything else is root-only including the #126 `TestAuthenticate`. Also records why `PreviewDetectFrame` and `TestAuthenticate` are spelled out explicitly rather than left to the root-only catch-all. **A5 is untouched.**

Also corrected the book's shorter user-facing copy, which named only `Enroll` and `Shutdown` as root-only — reading as though a `facelock`-group member may call the rest, including the preview methods.

### 10. Camera-tier test broken since N13 (MEDIUM, test) — ⚠️ UNVERIFIED BY EXECUTION

`check_preview_detect_frame_stripped` required `method return` in the dbus-send output, but `PreviewDetectFrame` became root-only, so a non-root caller gets `AccessDenied`, dbus-send exits non-zero, and the guard fails before it ever inspects the reply.

Its intent — a non-root caller must obtain no imagery — is still exactly right; only the mechanism changed, from a stripped reply to a denial before the method reaches the camera. It now asserts the denial: non-zero exit, `AccessDenied` in the output, and still no frame bytes in the error reply. The dbus-send invocation and the `set +e` / `rc` / `set -e` shape mirror `check_own_denied` in the same file.

**⚠️ This change is unverified by execution.** That tier needs a camera and ONNX models, so it cannot run in CI, in a container, or here. `bash -n` passes; the assertion logic has not been observed against a live daemon.

---

## Gates

All three run locally on the final tree.

- `just check` — **exit 0** (test, lint, fmt-check, audit, check-pam-standalone). Full workspace suite green; `cargo clippy --workspace -- -D warnings` clean. The "2 allowed warnings" line is cargo-audit's pre-existing allowlist, not new.
- `just test-arch-pam` — **22 passed, 0 failed** (unchanged)
- `just test-arch-layout` — **21 passed, 0 failed** (unchanged)

Each of the three code commits was also built in isolation to confirm the history is bisectable.

## Notes

- No sibling-agent files were touched. Two items produced findings in files owned by other agents this wave (`MockCamera::is_dark` thresholds, `direct.rs:67`); both are reported above rather than changed.
- Commits are signed (verified in the raw objects: `git cat-file commit <sha> | grep '^gpgsig'` shows an SSH signature on all five). Note that `git log --show-signature` and `%G?` report "No signature" on this machine because `gpg.ssh.allowedSignersFile` is unset — that is a verification-display gap, not an absence of signatures. Nothing was bypassed; no `--no-gpg-sign` was passed.

### DIFF
--- a/book/src/security.md
+++ b/book/src/security.md
@@ -112,7 +112,7 @@ For high-security deployments, embeddings can be encrypted with AES-256-GCM usin
 
 The D-Bus system bus policy (`/usr/share/dbus-1/system.d/org.facelock.Daemon.conf`) restricts which users and groups can own the bus name and invoke methods. Only root and members of the `facelock` group are granted access. (`/etc/dbus-1/system.d/` is the admin-override location for local customization.)
 
-Beyond bus-level access, the daemon enforces per-method UID authorization. Before executing any method, the daemon calls `GetConnectionUnixUser` to verify the caller's UID. `Authenticate` allows the caller's own UID. `Enroll` and `Shutdown` are restricted to root (UID 0) only. This prevents a `facelock` group member from enrolling faces or shutting down the daemon without root privileges.
+Beyond bus-level access, the daemon enforces per-method UID authorization. Before executing any method, the daemon calls `GetConnectionUnixUser` to verify the caller's UID. `Authenticate` allows the caller's own UID — a user must be able to request authentication for themselves, since screen lockers run their PAM stack as that user. Every other method, including `Enroll`, `Shutdown`, and the preview methods, is restricted to root (UID 0). This prevents a `facelock` group member from enrolling faces, pulling camera frames, or shutting down the daemon without root privileges.
 
 #### B. D-Bus Message Size Limits (Required)
 
--- a/config/quirks.d/00-defaults.toml
+++ b/config/quirks.d/00-defaults.toml
@@ -9,6 +9,12 @@
 #   name_pattern            - glob-like pattern matched against the WHOLE device
 #                             name (fallback). NOT regex: only (?i) and .* are
 #                             supported, and the pattern is anchored (#99).
+#                             Because the matcher has no alternation, a token
+#                             that can appear either at the start of a card
+#                             string or in the middle of one needs TWO entries:
+#                             a start-anchored form and a `.* `-prefixed,
+#                             space-bounded form. That is why the name entries
+#                             below come in pairs.
 #   force_ir                - treat this device as an IR camera regardless of format detection
 #   emitter_xu_guid         - UVC Extension Unit GUID for IR emitter control
 #   emitter_xu_selector     - UVC XU control selector for emitter toggle
@@ -45,18 +51,24 @@ notes = "Intel RealSense D455 IR module"
 
 # --- Lenovo ThinkPad IR cameras ---
 
-# Anchored (#99): name_pattern now matches the WHOLE device name, not a
-# substring of it, so this requires the name to literally START with
-# "IR Camera" (a trailing wildcard still allows a model-number suffix, e.g.
-# "IR Camera 5MP"). It intentionally does NOT match names where "IR Camera"
-# isn't the leading text (e.g. "Integrated IR Camera") — add a corrected,
-# still-anchored pattern for your exact device string (see `facelock
-# list-devices`) rather than reintroducing a bare leading wildcard here.
+# Anchored (#99): name_pattern matches the WHOLE device name, not a substring
+# of it. The first entry covers card strings that literally START with "IR
+# Camera" (a trailing wildcard still allows a model-number suffix, e.g. "IR
+# Camera 5MP"); the second covers the far more common ThinkPad spelling, where
+# "IR Camera" sits mid-string ("Integrated IR Camera"). The leading `.* ` is
+# space-bounded, so it still rejects a name that merely CONTAINS the letters
+# "ir" ("Sirius Camera", "AIR-Cam") — a bare leading `.*` would not.
 [[quirk]]
 name_pattern = "(?i)ir camera.*"
 force_ir = true
 warmup_frames = 8
 notes = "ThinkPad integrated IR camera (various models)"
+
+[[quirk]]
+name_pattern = "(?i).* ir camera.*"
+force_ir = true
+warmup_frames = 8
+notes = "ThinkPad integrated IR camera, vendor-prefixed name (various models)"
 
 # --- Logitech BRIO ---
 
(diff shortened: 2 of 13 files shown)


tyvsmith:

PR #132, opened 2026-08-14
polish(auth,daemon): callee-side wipe, honest seams, exhaustive matches (review polish P2)

tyvsmith:
Fourteen reviewer-identified defects in the authentication path and daemon, from the principal-architect review of PRs #107–#122, plus reconciliation items handed over by the hygiene package. No authentication decision changes: item 1 changes memory hygiene, item 10 changes a failure response. The two frozen protocol strings PAM substring-matches ("rate limited", "IR camera required") are byte-identical, the D-Bus wire is unchanged (`TestAuthenticate` from #126 and the `-4` model_id sentinel from #129 both intact), and `is-enrolled` is untouched.

## 1. The D11 zeroization contract was caller-side convention, implemented three ways

`authenticate_with_embeddings` took `&[(u32, FaceEmbedding)]` and did `.to_vec()` internally, so wiping the caller's plaintext was the caller's job. The wrapper that enforced it (`direct::authenticate_and_wipe`) lived in facelock-cli, which the daemon cannot depend on, so `handler.rs` carried an inline copy of the same wipe — one rule, two implementations. The `to_vec()` also created a third simultaneous plaintext copy purely to keep the signature immutable.

The parameter is now `&mut` and the callee owns the wipe. `authenticate_and_wipe`, the handler's inline duplicate, the `to_vec()` copy and the private `authenticate_inner` (merged into the public entry point) are all gone. Every caller — daemon handler, `facelock auth`, direct mode — hands over a buffer and gets it back zeroized.

**Pinning tests:** `auth_loop_wipes_the_callers_embeddings` and `auth_loop_wipes_the_callers_embeddings_on_failure` (`crates/facelock-daemon/tests/daemon_integration.rs`), covering both exit shapes — success returns from inside the loop, failure falls out of the bottom. Moved up from facelock-cli, where it could only cover the two CLI callers.

The signature change surfaced one real buffer reuse: `device_mismatch_never_reaches_success` ran a second auth against the same `stored`, which is now wiped by the first call. It rebuilds the compare set, which is what a caller must do.

## 2. No unwind path wiped anything (done, not deferred)

Both plaintext sets are held in a `Wiped` drop guard, so a panic wipes them too; the per-return-site `zeroize_stored_embeddings` and `variance_window.zeroize_all()` calls it replaces could not. The variance window and captured `Frame`s already zeroized on drop, so nothing else needed restructuring. `zeroize`'s own `Zeroizing` does not fit: it requires `T: Zeroize`, which the `(u32, FaceEmbedding)` tuple does not implement.

## 3. `validate_auth_bin` kept dead input validation and config-era naming

`AUTH_BIN` has been a compiled-in absolute constant since #109, so the empty-string and relative-path branches were unreachable — yet their messages and the syslog label still described a configurable `auth_bin`, and two tests pinned the dead branches. Renamed to `verify_binary_trust`; unreachable branches and their tests dropped; errors and the syslog label now say `untrusted_auth_binary`.

The runtime half is kept intact — canonicalize, then require the binary and every ancestor directory to be root-owned and non-group/world-writable. `canonicalize` returns an absolute path whatever it is given, so that walk establishes trust without the removed pre-check.

**Pinning tests:** `verify_binary_trust_rejects_missing_paths` (an absent binary is a real untrusted case) plus the two `validate_root_owned_nonwritable` tests. Dependency constraint honored: nothing added, `cargo build -p pam-facelock` clean, guard reports 57 crates.

## 4. Production bypassed the service constructor

`run_dbus_server` built `FacelockService` with a struct literal while `FacelockService::new` existed for tests, so an invariant added to the constructor would have missed the only instance that authenticates anyone. It now calls `new` and clones the handler and last-activity handles back off the service.

## 5. `unwrap()` on a lock in a library crate

The config-mtime lock in `maybe_reload_handler` used `unwrap()`, against the project rule and against the two sibling lock sites in the same function. A poisoned lock now reads as "no reload" — keep serving with the handler already built.

## 6. Startup and reload could disagree about the config file

`build_handler` took an explicit `config_path` at startup while the reload recipe passed `None` → `Config::load()`, and the mtime watch stats `paths::config_path()`. These agreed only because `main` also installs a process-level override. The redundant parameter is gone; `--config` still works through the override that was already doing the work.

## 7. Stale doc reference — already fixed upstream

The cited comment ("The D-Bus layer (commands/daemon.rs) passes false…") no longer exists anywhere in facelock-daemon; #126's `AuthIntent` rewrite replaced that whole doc block. No change needed.

## 8. The authorization matrix's completeness was convention

Adding a `Method` variant forced updating `name()` but not `Method::ALL`, which the matrix tests iterate — so the test comment claiming "a new method cannot be added without landing in the matrix" was aspiration. A `declare_methods!` macro now generates the enum, `ALL` and `name()` from one list, making the claim true by construction and deleting the duplicated name table. Drift here could only ever under-test (`scope()`'s catch-all keeps an unlisted method root-only), so this is hygiene, not a hole.

## 9. Wildcard arms defeated the exhaustiveness enum dispatch earns

All eight `Backend` operation methods matched `BackendKind::Daemon => …, _ => …`, so a fourth kind would have silently inherited direct-transport behavior in eight places instead of producing eight compile errors. All eight now spell out both direct variants, as `caps()` already did — as does `is_direct`, whose `!matches!` was the same wildcard in miniature.

## 10. A seam method whose contract depended on call ordering

`clear_models` errored on `StoreError::Absent` because "callers check `has_models` first", while the sibling `remove_model` answers `Some(false)` for the same state. Absent now means `Some(0)`, routed through `direct_read` so it still never materializes the database it reports empty (C7). An unreadable store remains an error on every operation.

**Pinning test:** `absent_store_reads_are_benign_and_create_nothing` now asserts `clear_models → Some(0)` beside the sibling reads it agrees with, and still asserts no database was created.

## 11. Init-system-specific hints in user-facing messages

Both the daemon-unreachable fallback and the preview-unreachable message hardcoded `sudo systemctl start facelock-daemon`, though this repo also ships an OpenRC service; the fallback prints on every backend-using invocation in a degraded state. Both now say to start the facelock-daemon service without naming an init system.

These strings live in `crates/facelock-cli/src/message/access.rs`, the current location after #130's per-domain split. `po/facelock.pot` regenerated with `just pot`; `po/pam_facelock.pot` picks up unrelated line-number drift in the same pass, having not been regenerated since lib.rs last moved.

## 12 & 13. Reconciliation handed over by the hygiene package

Both are in files this branch owns, so they landed here rather than on `polish/p3-hygiene`.

**`PamNotificationConfig` had two lists of defaults.** Per-field `#[serde(default = ...)]` alongside a manual `Default` impl — the exact shape that already shipped a drift bug in this struct (the mode enum's `#[default]` said `Both` while the struct's said `Terminal`, undetected because `terminal()` treats them identically). The two agree today, so moving `#[serde(default)]` to the container and deleting the per-field ones is behavior-preserving: "section present, key omitted" and "section absent" are now identical by construction. `PamNotificationMode`'s `Default` impl is deleted rather than left unused — it was the second answer that drifted, and nothing called it. `default_true` stays, since `PamSecurityConfig` still uses it.

The hygiene package made the same change to facelock-core's `NotificationConfig`. The two are hand-kept mirrors (the PAM dependency ceiling forbids sharing the type) and must keep agreeing.

**Pinning tests:** both existing drift pins pass unchanged and now guard the container mechanism — `notification_section_present_with_keys_omitted_equals_absent` and `shipped_template_notification_settings_equal_defaults`.

**`server.rs` called desktop delivery "fire-and-forget".** It is not: `notifications.rs` runs the helper with `Command::output()`, which waits for the child, and the call sits inside the `spawn_blocking` closure before the reply is constructed — so an auth reply waits on it. The comment now says so and states the real contract (delivery must not *fail* an auth; staying cheap is the implementation's obligation, not a guarantee of the call site). Delivery is not restructured.

## 14. Daemon-crate lints under the `--all-targets` gate

Added to this package by the orchestrator after the hygiene package tightened `just lint` to `cargo clippy --workspace --all-targets -- -D warnings`, which makes test code visible to clippy for the first time.

Surveyed per-package rather than workspace-wide, because `facelock-store`'s two non-octal-permissions **errors** abort that crate's test build and truncate a workspace run before it reaches later crates — a workspace-wide green is not evidence here.

`cargo clippy -p facelock-daemon --all-targets -- -D warnings` reports **nine** hits, all `clippy::type_complexity`, all the same hand-spelled box:

```
Box<dyn Fn(&Config) -> Result<MockCamera, String> + Send + Sync>
```

at `daemon_integration.rs` 338/384/626/701/727/786/874, `server_authz.rs:78`, `test_authenticate_ssh_gate.rs:64`. There are no `needless_range_loop`, `redundant_closure`, manual-fill or needless-borrow lints in this crate — those live in crates the hygiene package owns.

All nine are now a `MockCameraFactory` alias. Type annotations only: no test logic, no assertions, no test names, no production code. The alias is defined once per test file because each integration test file is its own crate; sharing it would mean exporting a test-only type from the daemon's production API or adding it to `facelock-test-support`, which this package does not own. `server_authz.rs` already had a local `MockService` alias, so this matches the file's idiom.

The name is deliberately the one the hygiene package intended to export from `facelock-test-support` before these files were reassigned here — so if that export lands later, resolution is to delete three `type` lines and add one `use`, with every annotation already correct.

`cargo clippy -p facelock-daemon --all-targets -- -D warnings` is clean; all 130 daemon tests pass.

## Gates

Re-run in full after every added item:

- `just check` — pass (test, clippy `-D warnings`, fmt, audit, pam standalone guard at 57 crates)
- `just test-arch-pam` — 22 passed, 0 failed
- `just test-arch-layout` — 21 passed, 0 failed

Review finding: REV-109, REV-121, REV-120, REV-108.

### DIFF
--- a/crates/facelock-cli/src/backend.rs
+++ b/crates/facelock-cli/src/backend.rs
@@ -60,7 +60,10 @@ pub enum BackendKind {
 
 impl BackendKind {
     pub fn is_direct(self) -> bool {
-        !matches!(self, BackendKind::Daemon)
+        match self {
+            BackendKind::Daemon => false,
+            BackendKind::DirectByConfig | BackendKind::DirectByFallback => true,
+        }
     }
 }
 
@@ -157,11 +160,13 @@ impl<'a> Backend<'a> {
     pub fn has_models(&self, user: &str) -> anyhow::Result<bool> {
         match self.kind {
             BackendKind::Daemon => Ok(!self.daemon_list_models(user)?.is_empty()),
-            _ => self.direct_read(false, |store| {
-                store
-                    .has_models(user)
-                    .map_err(|e| anyhow::anyhow!("storage error: {e}"))
-            }),
+            BackendKind::DirectByConfig | BackendKind::DirectByFallback => {
+                self.direct_read(false, |store| {
+                    store
+                        .has_models(user)
+                        .map_err(|e| anyhow::anyhow!("storage error: {e}"))
+                })
+            }
         }
     }
 
@@ -172,11 +177,13 @@ impl<'a> Backend<'a> {
                 .daemon_list_models(user)?
                 .iter()
                 .any(|model| model.embedder_model == embedder)),
-            _ => self.direct_read(false, |store| {
-                store
-                    .has_models_for_embedder(user, embedder)
-                    .map_err(|e| anyhow::anyhow!("storage error: {e}"))
-            }),
+            BackendKind::DirectByConfig | BackendKind::DirectByFallback => {
+                self.direct_read(false, |store| {
+                    store
+                        .has_models_for_embedder(user, embedder)
+                        .map_err(|e| anyhow::anyhow!("storage error: {e}"))
+                })
+            }
         }
     }
 
@@ -184,11 +191,13 @@ impl<'a> Backend<'a> {
     pub fn list_models(&self, user: &str) -> anyhow::Result<Vec<FaceModelInfo>> {
         match self.kind {
             BackendKind::Daemon => self.daemon_list_models(user),
-            _ => self.direct_read(Vec::new(), |store| {
-                store
-                    .list_models(user)
-                    .map_err(|e| anyhow::anyhow!("storage error: {e}"))
-            }),
+            BackendKind::DirectByConfig | BackendKind::DirectByFallback => {
+                self.direct_read(Vec::new(), |store| {
+                    store
+                        .list_models(user)
+                        .map_err(|e| anyhow::anyhow!("storage error: {e}"))
+                })
+            }
         }
     }
 
@@ -204,12 +213,14 @@ impl<'a> Backend<'a> {
             // An absent store provably holds nothing to remove: report "not
             // found" without materializing a database (the per-site version
             // used the creating opener here).
-            _ => self.direct_read(Some(false), |store| {
-                store
-                    .remove_model(user, model_id)
-                    .map(Some)
-                    .map_err(|e| anyhow::anyhow!("{e}"))
-            }),
+            BackendKind::DirectByConfig | BackendKind::DirectByFallback => {
+                self.direct_read(Some(false), |store| {
+                    store
+                        .remove_model(user, model_id)
+                        .map(Some)
+                        .map_err(|e| anyhow::anyhow!("{e}"))
+                })
+            }
         }
     }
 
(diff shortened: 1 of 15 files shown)


tyvsmith:

PR #133, opened 2026-08-14
polish(hygiene,tests): stricter lint gate, real coverage, shared fixtures (review polish P3)

tyvsmith:
Fifteen hygiene and test-quality defects from the principal-architect review of #107–#122 (plus one found during the fix waves). Behavior changes are confined to three deliberate ones, called out below; everything else is docs, tests, and pinning. No auth-semantics change, no D-Bus wire change, `is-enrolled` untouched.

> ### ⚠️ `just lint` is RED on this branch, deliberately
>
> Exactly nine `clippy::type_complexity` warnings remain, all in `crates/facelock-daemon/tests/**`. That crate is owned by the concurrent auth-path change (#132) and the orchestrator assigned its lints there exclusively; this branch is byte-identical to the merge base under `crates/facelock-daemon/`.
>
> This is expected and correct. The gate is right, the branch that owns those files is fixing them in parallel, and **the combination is green**. Weakening the recipe to make this branch green on its own would defeat the entire point of the change, so it has not been done. Merge gating belongs on the combination, not on this PR alone.
>
> Every other crate is at zero — per-package sweep, because the workspace-wide run used to truncate (see below):
>
> `core 0 · camera 0 · face 0 · store 0 · cli 0 · bench 0 · pam 0 · tpm 0 · polkit 0 · test-support 0 · **daemon 9**`

## The gate, and why it is first

**1. `just lint` omitted `--all-targets`, so deny-by-default lints in test code never reached the standing gate.** — *Review finding: REV-114*

`cargo clippy --workspace` checks only default targets. Every lint inside `#[cfg(test)]`, and in the bench binaries, was invisible. That gap matters disproportionately here because file modes *are* a security contract (0600 database, 0710 state dir), and `non_octal_unix_permissions` is exactly the lint that catches `from_mode(600)` — which means `0o1130`, not `0o600`.

There is no permission bug today: the two offenders were `from_mode(0)`, where decimal and octal agree. The point is that the gate could not have told us either way.

`justfile` and both CI clippy steps now run `cargo clippy --workspace --all-targets -- -D warnings`. That exposed **25 warnings + 2 errors**:

| Lint | N | Where | Fix | Owner |
|---|---|---|---|---|
| `non_octal_unix_permissions` (**error**) | 2 | store/db.rs | `from_mode(0o000)` | fixed here |
| `needless_range_loop` | 9 | store/db.rs ×2, store_integration.rs, core/types.rs ×2, cli/bench.rs ×2, bench/main.rs ×2 | `zip`+`enumerate`, or slice `fill` where the loop only wrote | fixed here |
| `manual_slice_fill` | 2 | core/types.rs | `fill` | fixed here |
| `redundant_closure` | 3 | cli/setup.rs | pass the fn item | fixed here |
| `needless_borrow` | 1 | tpm/sealing.rs | drop the `&` | fixed here |
| `type_complexity` | 1 | store/error.rs | `IsExpectedClass` alias | fixed here |
| `type_complexity` | 9 | daemon tests | `MockCameraFactory` alias | **#132** |

> **Methodology note, worth keeping.** The two `non_octal_unix_permissions` findings are *errors*, not warnings — they abort the `facelock-store` lib-test build, which **truncates a workspace-wide clippy run and hides every later package**. Any survey taken from one workspace invocation undercounts. Take counts per-package. (Those two are fixed here, so the truncation is gone, but the method still stands.)

All fixes are test or bench code. No test logic was restructured to satisfy a style lint.

## Tests that looked like they covered something

**2. A guard test silently stopped covering its branch.** — *REV-115*

`zero_models_proceeds` (cli/setup.rs) used to exercise "store opens, zero models" because the old create-based open materialised an empty database as a side effect. After #115 retyped the guard to `open_store_existing` with `StoreError::Absent → Ok(())`, it never created the file — so it landed on the Absent early return and became a duplicate of `absent_database_proceeds_without_creating_one`. The real arm (a database with zero rows: the post-`facelock clear` state) had no test.

Fixed: the test now creates and drops a real `FaceStore`, reaching `has_any_models() == Ok(false)`. The Absent test is unchanged. The stale doc comment is corrected.

**3. Schema-level fault injection was bespoke arcana duplicated across two crates.** — *REV-115*

cli/setup.rs did `writable_schema` rootpage repointing; the daemon's integration tests did a `DROP COLUMN`; each carried a private `rusqlite` dev-dependency. Both were correct and failed loudly, but each encodes non-obvious coupling — migration version gating, the exact column set a query selects, SQLite's autoindex naming — so a schema change breaks them in two places, each needing the same re-derivation.

Both are now `facelock-test-support::schema_faults::{break_face_models_table, drop_embedder_model_column}`, with that coupling documented *and asserted* (an injector that silently fails to inject would turn its test into a test of the happy path). No store trait seam — reviewers judged that worse than the raw SQL.

> **Half of this item is cross-branch.** `break_face_models_table` is converted and in use here. `drop_embedder_model_column`'s only caller is the daemon integration test, which is #132's file — so it is exported and documented here, and the one-line call-site conversion (plus dropping the daemon's now-redundant `rusqlite` dev-dep) lands there. Signature and call site were handed over explicitly. It is kept exported rather than deleted so the item lands half-*applied*, not half-*removed*.

**4. A test mutated process-global env with only a comment protecting it.** — *REV-115*

`shared_resolver_honors_sudo_user` called `unsafe set_var("SUDO_USER", …)` in a threaded harness, with a SAFETY note asserting no other test in the binary reads SUDO_USER — true when written, enforced by nothing. The resolution order is now `ipc_client::resolve_user_from_env`, which takes the environment as a parameter; the test injects. `resolve_user` is a one-line delegation.

**9. A template test skipped lines silently.** — *REV-122*

`every_documented_example_default_round_trips` embeds a hand-tuned prose-vs-assignment classifier and only asserts on lines it *did* uncomment. Anything it declined contributed no coverage and no signal — which is where a multi-line example would land. It now records assignment-shaped lines it skipped and pins the count (two, both prose), so an example the classifier cannot read fails the build.

## Behavior changes (three, all deliberate)

**5. A pure-read path created an empty database.** *(behavior change)* — *REV-116*

`preview --text-only` in direct mode loads embeddings to show live similarity and never writes, yet called the create-based `open_store` — materialising an empty database on a fresh install, the exact "silent lie about a store of biometric templates" `open_existing`'s doc warns about. Now uses `open_store_existing` and maps `Absent` to an empty set plus the catalog's "no models enrolled" notice; every face simply comes back unrecognized.

**12. The notification default-drift class is now impossible, not just pinned.** *(no observable change — the two mechanisms already agreed)* — *REV-118*

`NotificationConfig` carried per-field `#[serde(default = …)]` *and* a manual `impl Default` — two mechanisms filling the same struct, agreeing only by hand. They did not agree once: `notify_on_failure` read `true` from serde and `false` from `Default`, so a shipped release flipped the effective value depending on whether `[notification]` appeared in the file. The template has exactly that shape (active header, all keys commented).

Container-level `#[serde(default)]` plus removal of the per-field defaults makes the manual `Default` the single source, so "section present, key omitted" ≡ "section absent" by construction. The existing pin stays as a regression guard.

That left `NotificationMode`'s own `Default` unused — a second answer to "what is the default mode", and the impl that drifted to `Both` last time. Deleting it was the alternative, but its two sibling enums in the same file (`SnapshotMode`, `EncryptionMethod`) still have live `Default`s via their own field-level defaults, so removing just this one reads as an oversight rather than a decision. It is pinned equal to `NotificationConfig::default().mode` instead, which also covers what deletion could not: if `#[serde(default)]` is ever re-added to the `mode` field, the enum's answer becomes authoritative again and the pin catches it disagreeing. (#132 made the mirrored PAM-side change and chose deletion there; the divergence is deliberate and reasoned on both sides.)

**14. The oneshot-fallback verdict overstated what it checked.** *(behavior change: rendered text)* — *REV-116*

`usable()` checked binary + models + database, then rendered "usable (root-invoked PAM can authenticate…)". By the PR's own "fails one user vs fails the mechanism" principle that is not enough: a missing camera or key material fails the mechanism too, and both already live in `Health`. And presence ≠ loadability — a SHA-mismatched model passes.

Folding two more sections in would mean two places to keep true, so the claim is narrowed to the evidence instead: `usable()` → `prerequisites_present()`, and the text becomes *"prerequisites present (binary, models and database in place for daemon-less auth)"* / *"prerequisites missing (…)"*. What is deliberately not covered, and where it is reported instead, is documented on the type.

> **Golden test updated deliberately.** Two literals in `commands/status.rs` changed: the `[ok]` line in `healthy_report_renders_byte_exact`, and the `[!!]` line in the per-status table. Both are the strings above. **No container test greps either one** (verified against `test/`). `po/facelock.pot` is regenerated for the two changed msgids plus one added; `po/pam_facelock.pot` is deliberately left alone — its only diff was pre-existing line-reference drift in a crate this branch does not touch.

## Smaller corrections

**6. Error detail dropped the SQLite code whenever a message existed.** — *REV-113* — rusqlite renders `SqliteFailure(_, Some(msg))` as the message alone, and the message is often the *less* useful half: ~30 `SQLITE_IOERR_*` codes all read "disk I/O error", and which one it was is what says full disk vs bad sector vs broken lock. `classify` now appends the extended code. Pinned by `classify_keeps_the_result_code_beside_the_message`.

**7. A path was recovered instead of threaded.** — *REV-113* — `migration_err` rebuilt the path from `conn.path()`, falling back to a literal `":memory:"` — inventing an answer — only because `run_migrations` didn't take the path both callers already hold. Now it does; the fallback is gone, and the signature drops to `pub(crate)` (zero external callers).

**8. A conformance test compared configs by `Debug` string.** — *REV-122* — sound today (no custom `Debug` impls, no `HashMap`s) but silently degrading the moment a field gains redaction or unordered iteration. `Config` and its twelve nested sections now derive `PartialEq`; `Config` also derives `Default`, composing each section's own. That generalizes item 12's guard: `empty_document_parses_to_default` asserts `Config::parse("") == Config::default()`, so *any* section whose field default disagrees with its `Default` impl fails in CI.

**10. The parse-once pin had an obvious bypass.** — *REV-116* — it counted `Config::load()` per file, but `Config::load_from(paths::config_path())` walks around it — and that is the natural copy-paste, since `ConfigLoad::read_from` models the shape. Two new halves: composing both on one line is allowed nowhere; naming the global config path at all is allowed only in the four files that read or write the file itself (per-file, not per-count, so setup's config writes can churn freely while a new file reaching for the global path stays a decision).

**11. A hardcoded provider list re-created the drift the PR deleted.** — *REV-122* — the hint spelled "cpu, cuda, rocm, openvino" as a literal while facelock-face derives the list from `ProviderKind::ALL`. The literal must stay (it is a translated sentence; an assembled one cannot be localized), so `ALL` is now public with an `all_names()` accessor, and `unknown_provider_hint_lists_every_provider` holds the literal to the enum.

**13. A trait doc promised more than the implementation delivers.** — *REV-118* — `Notifier` said implementations "must never block or fail the auth flow", but desktop delivery runs `setpriv` + `notify-send` through `Command::output()`, which waits for the child, before the D-Bus reply returns. The doc now states the contract that holds — delivery must never *fail* auth; errors are logged and swallowed, and `notify` returns nothing so no caller can branch on one — and the "fire-and-forget" claim in cli/notifications.rs is corrected. Delivery is unchanged. (The matching daemon-side comment at `server.rs:591` was handed to #132 and is fixed there.)

**15. Two more couplings in health.rs.** — *REV-116* — `AUTH_BIN` is genuinely duplicated from pam-facelock (which may not depend on any facelock crate) with only a comment holding the copies together; a divergence would be silent, with `status` reporting on a binary PAM never spawns. `auth_bin_matches_the_path_pam_actually_spawns` now reads PAM's declaration and compares. And `CONFIG_NOT_AVAILABLE` was authored English embedded inside the localized "cannot determine: {why}" sentence — it is now a message-catalog variant.

## Coordination

- **`crates/facelock-daemon` is untouched** — byte-identical to the merge base. Its nine lints, the item-3 call site, and the item-13 comment all belong to #132.
- **`crates/facelock-core/src/types.rs`** has four style-lint fixes (in the `cosine_similarity_*` tests) despite belonging to #131's author, who confirmed they are not mid-edit and to take these hunks. They sit ~400 lines from that branch's changes; a clean auto-merge is expected.
- **`MockCameraFactory`** is exported from facelock-test-support with no in-tree caller. #132 mirrors it under an identical local name in three test files, specifically so the merge collapses to deleting three `type` lines and adding a `use`. A crate-root re-export means either spelling works.

## Merging: `po/facelock.pot` needs regenerating, not resolving

This branch and #132 both regenerate the gettext template, so they conflict in `po/facelock.pot`. **Do not accept any resolution of that file — including the one git offers you without asking.**

`git rerere` is enabled here and holds a cached resolution for this exact conflict. The merge does not stop: it prints `Staged 'po/facelock.pot' using previous resolution.` and leaves zero conflict markers. That resolution is **wrong** — it keeps this branch's side of two hunks and drops #132's rewording, so the template ends up documenting `"Start it with: sudo systemctl start facelock-daemon"` while the merged code says `"Start the facelock-daemon service to use it."`. Two of #132's user-facing strings would ship with no catalog entry.

The `msgfmt --check-format` gate from #130 does **not** catch this. Verified against both the correct and the bad template: both pass. It validates the placeholder contract, not msgid-versus-source agreement.

Verified resolution:

```
git rerere forget po/facelock.pot   # otherwise it resolves before you can choose
git checkout --ours po/facelock.pot # either side; the content is about to be replaced
just pot                            # regenerate from the merged sources
```

That yields a template with all four contested strings correct and the stale wording gone, and it validates. The file is a build artifact; the only correct version is the one generated from the merged tree.

## Gates

- `cargo test --workspace` — **29 test targets, 0 failures**
- `just check` — **RED, and only on the nine daemon lints above.** Every other package is zero under `cargo clippy -p <pkg> --all-targets -- -D warnings`. Test, fmt, audit and the PAM standalone-surface guard all pass.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,11 +42,14 @@ jobs:
       - name: Build workspace with TPM
         run: cargo build --workspace --features tpm
 
+      # --all-targets so test/bench code is linted too; without it a
+      # deny-by-default lint in a test never fails CI. Keep in sync with the
+      # justfile `lint` recipe, which is the same command.
       - name: Run clippy
-        run: cargo clippy --workspace -- -D warnings
+        run: cargo clippy --workspace --all-targets -- -D warnings
 
       - name: Run clippy with TPM
-        run: cargo clippy --workspace --features tpm -- -D warnings
+        run: cargo clippy --workspace --all-targets --features tpm -- -D warnings
 
       - name: Run tests
         run: cargo test --workspace
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1082,6 +1082,7 @@ dependencies = [
  "image",
  "libc",
  "nix",
+ "rusqlite",
  "serde",
  "serde_json",
  "signal-hook",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,7 +1038,6 @@ dependencies = [
  "libc",
  "nix",
  "reqwest",
- "rusqlite",
  "serde",
  "serde_json",
  "sha2",
@@ -1083,7 +1082,6 @@ dependencies = [
  "image",
  "libc",
  "nix",
- "rusqlite",
  "serde",
  "serde_json",
  "signal-hook",
@@ -1138,6 +1136,7 @@ name = "facelock-test-support"
 version = "0.1.4"
 dependencies = [
  "facelock-core",
+ "rusqlite",
 ]
 
 [[package]]
--- a/crates/facelock-bench/src/main.rs
+++ b/crates/facelock-bench/src/main.rs
@@ -921,12 +921,8 @@ mod tests {
         let mut emb_a = [0.0f32; 512];
         let mut emb_b = [0.0f32; 512];
         // Orthogonal embeddings
-        for i in 0..256 {
-            emb_a[i] = 1.0 / (256.0f32).sqrt();
-        }
-        for i in 256..512 {
-            emb_b[i] = 1.0 / (256.0f32).sqrt();
-        }
+        emb_a[..256].fill(1.0 / (256.0f32).sqrt());
+        emb_b[256..].fill(1.0 / (256.0f32).sqrt());
         let faces = vec![(
             facelock_core::types::Detection {
                 bbox: facelock_core::types::BoundingBox {
--- a/crates/facelock-cli/Cargo.toml
+++ b/crates/facelock-cli/Cargo.toml
@@ -51,9 +51,9 @@ bytemuck = "1"
 
 [dev-dependencies]
 tempfile = "3"
+# Schema-level failure injection lives in facelock-test-support::schema_faults,
+# which owns the rusqlite dependency it needs.
 facelock-test-support = { path = "../facelock-test-support" }
-# Raw SQL access for schema-level failure injection in tests.
-rusqlite = { version = "0.40", features = ["bundled"] }
 
 [features]
 default = ["wayland"]
--- a/crates/facelock-cli/src/commands/bench.rs
+++ b/crates/facelock-cli/src/commands/bench.rs
@@ -796,12 +796,8 @@ mod tests {
     fn test_find_best_match_below_threshold() {
         let mut emb_a = [0.0f32; 512];
         let mut emb_b = [0.0f32; 512];
-        for i in 0..256 {
-            emb_a[i] = 1.0 / (256.0f32).sqrt();
-        }
-        for i in 256..512 {
-            emb_b[i] = 1.0 / (256.0f32).sqrt();
-        }
+        emb_a[..256].fill(1.0 / (256.0f32).sqrt());
+        emb_b[256..].fill(1.0 / (256.0f32).sqrt());
         let faces = vec![(
             facelock_core::types::Detection {
                 bbox: facelock_core::types::BoundingBox {
--- a/crates/facelock-cli/src/commands/preview/mod.rs
+++ b/crates/facelock-cli/src/commands/preview/mod.rs
@@ -104,16 +104,24 @@ mod tests {
     /// getpwuid-only version lacked.
     #[test]
     fn shared_resolver_honors_sudo_user() {
-        // SAFETY: process-global env mutation; no other test in this binary
-        // asserts on SUDO_USER (checked), and the one concurrent reader
-        // (`resolve_user_no_flag_falls_through`) only requires a non-empty
-        // result.
-        unsafe { std::env::set_var("SUDO_USER", "preview-c5-alice") };
-        let resolved = crate::ipc_client::resolve_user(None);
-        unsafe { std::env::remove_var("SUDO_USER") };
-        assert_eq!(resolved, "preview-c5-alice");
+        // The environment is injected rather than set: `set_var` is global to
+        // the test binary and cargo runs these threaded, so the real thing
+        // would be safe only by convention.
+        let env = |key: &str| match key {
+            "SUDO_USER" => Some("preview-c5-alice".to_string()),
+            "USER" => Some("root".to_string()),
+            _ => None,
+        };
+
+        assert_eq!(
+            crate::ipc_client::resolve_user_from_env(None, env),
+            "preview-c5-alice"
+        );
 
         // The explicit flag still wins over the environment.
-        assert_eq!(crate::ipc_client::resolve_user(Some("bob")), "bob");
+        assert_eq!(
+            crate::ipc_client::resolve_user_from_env(Some("bob"), env),
+            "bob"
+        );
     }
 }
--- a/crates/facelock-cli/src/commands/preview/text_only.rs
+++ b/crates/facelock-cli/src/commands/preview/text_only.rs
@@ -4,6 +4,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::time::Instant;
 
 use crate::ipc_client;
+use crate::message::{FaceMessage, Terminal};
 
 /// Run the text-only preview mode.
 ///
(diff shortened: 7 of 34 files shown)


tyvsmith:

PR #142, opened 2026-08-15
ci: gate on translation catalog freshness

tyvsmith:
## Summary

`po/facelock.pot` and `po/pam_facelock.pot` are generated from source by `just pot`. Nothing currently verifies the committed copies still match the code — `msgfmt --check-format` (wired into `just mo`) only validates the `{placeholder}` contract of an existing translation against its own msgid, not whether msgids still correspond to strings the code contains. A stale catalog passes every other gate silently. This already happened once: a merge conflict in `po/facelock.pot` was silently auto-resolved by `git rerere` using a stale cached resolution, dropping two reworded user-facing strings, while `just check` and both container tiers stayed green.

Adds a CI-only job (`catalog-check`) that runs `just pot && git diff --exit-code po/` in the same `archlinux:base-devel` container the other jobs use, failing the build on drift. Covers both templates in one job since `just pot` regenerates both.

**Deliberately not added to `just check`.** PR #130 made the `pot`/`mo` recipes opt-in and tolerant specifically so building, testing, and `just check` never require gettext. Wiring `just pot` into `just check` would silently reverse that and make `xgettext` a hard dependency for every contributor just to catch one class of bug. This job is CI-only.

## Verification

- Confirmed `just pot && git diff --exit-code po/` is currently clean on `main` (catalog matches source).
- Proved the gate can actually fail: temporarily reworded a message in `crates/facelock-cli/src/message/device.rs` without regenerating the catalog, ran the job's exact commands — `git diff --exit-code po/` exited 1 with the expected `msgid` diff. Reverted and reran — clean (exit 0).
- The two benign `unterminated string literal`/`unterminated character constant` warnings `just pot` emits from `crates/pam-facelock/src/lib.rs` go to stderr and do not affect `just pot`'s exit code or the diff; left untouched per the plan (they are pre-existing xgettext C-tokenizer noise on doc-comment apostrophes).
- `just check`: pass (exit 0).
- `just test-arch-pam`: 22 passed, 0 failed.
- `just test-arch-layout`: 21 passed, 0 failed.

## Test plan

- [x] `just check`
- [x] `just test-arch-pam` (22/22)
- [x] `just test-arch-layout` (21/21)
- [x] Manually proved the gate fails on a stale catalog and passes when clean
- [ ] CI itself runs the new `catalog-check` job on this PR (targets `main`, so it will)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,6 +169,41 @@ jobs:
           TCTI: swtpm:host=localhost,port=2321
         run: cargo test --workspace --features tpm
 
+  catalog-check:
+    name: Translation Catalog Freshness
+    runs-on: ubuntu-latest
+    container: archlinux:base-devel
+    steps:
+      - name: Install system dependencies
+        run: |
+          pacman -Syu --noconfirm \
+            git \
+            gettext \
+            just
+
+      - uses: actions/checkout@v7
+
+      # po/*.pot are GENERATED from source by `just pot`. Nothing else
+      # verifies the committed copies still match the code: `msgfmt
+      # --check-format` (see the `mo` recipe) only validates the
+      # `{placeholder}` contract of an existing translation against its own
+      # msgid — it says nothing about whether msgids still correspond to
+      # strings the code actually contains. A stale catalog (e.g. from a bad
+      # merge/rerere resolution — this has already happened once, see
+      # po/facelock.pot's history) passes every other gate silently.
+      # Regenerate and diff: any drift fails the build.
+      #
+      # Deliberately NOT part of `just check`: PR #130 made the pot/mo
+      # recipes opt-in and tolerant specifically so building, testing, and
+      # `just check` never require gettext. Wiring `just pot` into `just
+      # check` would silently reverse that and make xgettext a hard
+      # dependency for every contributor just to catch one class of bug.
+      # This job is CI-only for exactly that reason.
+      - name: Verify translation catalogs are up to date
+        run: |
+          just pot
+          git diff --exit-code po/
+
   container-pam-test:
     name: Container PAM Smoke Test
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -183,6 +183,22 @@ jobs:
 
       - uses: actions/checkout@v7
 
+      # This is the only job that runs git against the checkout itself (the
+      # others only run cargo/just), and inside a container git refuses it:
+      # the steps run as root while /__w is owned by the runner uid, which
+      # trips the dubious-ownership check. actions/checkout does add a
+      # safe.directory entry, but only under a temporary HOME it discards when
+      # the action finishes ("Temporarily overriding HOME=... before making
+      # global git config changes" in its log), so later steps never see it.
+      #
+      # The symptom is not the usual "detected dubious ownership" fatal:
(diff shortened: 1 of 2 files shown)


tyvsmith:
## Follow-up: the gate never ran, and two things had to be fixed before it could

Brought up to date with `main` (through #145) and fixed the job. The first CI run of this job died with exit 129 and git's usage text, so it had never actually compared a catalog.

**1. `git diff` was rejected by git, not by the catalog.** This is the only job that runs `git` against the checkout — the others only run `cargo`/`just`. Inside the container the steps run as root while `/__w` is owned by the runner uid, so git rejects the repository as dubiously owned. `actions/checkout` does add a `safe.directory` entry, but under a temporary `HOME` it discards when the action finishes (`Temporarily overriding HOME=... before making global git config changes` in its log), so later steps never see it. It did not surface as the usual `detected dubious ownership` fatal because `git diff` supports `--no-index`: it sets the repository up *gently*, so a rejected repo degrades to "not a git repository", prints `warning: Not a git repository. Use --no-index ...` plus usage, and exits 129 without comparing anything. Fixed by adding the `safe.directory` entry in a step of our own.

**2. `POT-Creation-Date` is not reproducible in CI.** Latent second failure that would have bitten immediately after fixing (1). gettext 1.0's `xgettext` stamps that header with the *commit* timestamp of the newest input file when it can read the file's git history, and falls back to the file's **mtime** when it cannot. Verified locally: in a full clone, regenerating leaves the header untouched even days later and even after `touch`-ing the source; in a `--depth 1` clone it takes the mtime path. CI checks out with `fetch-depth: 1`, so it always stamps the moment the runner cloned — i.e. a fresh value on every run. That is why the check was clean locally for @tyvsmith but could never have been clean in CI. The diff now excludes that single header line (`-I '^"POT-Creation-Date: '`); msgids, source references and flags — what the gate is actually about — are still compared exactly. Verified both directions locally: a date-only change exits 0, a real msgid/location change still exits 1.

## First real finding: the committed PAM catalog was stale

With the gate finally running, `po/pam_facelock.pot` was out of date on `main`. #145 (camera lifecycle) added ~340 lines to `crates/pam-facelock/src/lib.rs` and moved every translated call site, so the `#:` source references pointed at the wrong lines (`lib.rs:991` → `1166`, etc.). Regenerated in its own commit. **No msgid changed** — both messages (`Identifying face...`, `Face recognized.`) and all four call sites are still extracted, so no user-facing string was lost. `po/facelock.pot` was already in sync.

Note that this is what the gate will do from now on: any PR that shifts line numbers in `crates/facelock-cli/src/message/*.rs` or `crates/pam-facelock/src/lib.rs` fails until the author reruns `just pot`. If that churn gets annoying, `--add-location=file` in the `pot` recipe would keep the file references but drop the line numbers.

## Verdict on the `xgettext` warnings: benign, catalog is complete

The `unterminated string literal` / `unterminated character constant` warnings are xgettext's C tokenizer meeting Rust: line 1057 is `fn pam_code_for_daemon_error(...) -> (libc::c_int, &'static str)` — the lifetime tick reads as an unterminated char constant — and the rest are `r#"` raw strings in `#[cfg(test)]` code. The scanner resynchronises at end of line, so nothing after them is swallowed: probed by inserting a `gettext("...")` call immediately after each of the nine warned lines and confirming all nine were extracted. All four real call sites appear in the catalog. Left as-is; a proper fix means a Rust-aware extractor, which is a separate change.

PR #145, opened 2026-08-15
Camera lifecycle: hold only after failure, cancellable auth, one-shot hygiene (ADR 008)

tyvsmith:
## Summary

Implements ADR 008 (`docs/adr/008-camera-lifecycle.md`): the camera is on only while an authentication is in progress or a retry is plausibly imminent. Success, cancel and error release it immediately; a no-match/timeout keeps it warm for `device.camera_release_secs` (now default 3, `0` = never; previously 5 after *every* request). Every in-flight request carries a cancel token set when the caller's bus connection disappears, on suspend/shutdown, or (one-shot) on SIGTERM. Warm reuse discards the stale V4L2 buffers first. One-shot loads the engine before opening the camera and dies with its PAM host. `recognition.no_face_timeout_secs` (default 2) ends attempts at an empty chair early, and such attempts no longer charge the rate limiter. No config migration.

Motivation: [omarchy#4982](https://github.com/basecamp/omarchy/discussions/4982) (IR LEDs strobe for the 5 s hold after every auth) and [omarchy PR #6863](https://github.com/basecamp/omarchy/pull/6863) (auto-start on lid/key, 3 × 250 ms retries, password typed in parallel, PAM aborted on unlock/lid-close).

## Commits (one per ADR §10 row; please review commit-by-commit, merge with a merge commit)

- `3158f37` docs(adr): ADR 008 — outcome-based camera hold and cancellable authentication
- `11bcb59` feat(daemon): hold the camera only after a failed attempt — `CameraLease` (acquire/finish/touch_preview/expire), outcome rule, `0` = never, default 3, 250 ms poll on an absolute deadline, `MMAP_BUFFERS − 1` stale discard on warm reuse, preview floor, docs/contracts/CHANGELOG
- `0ad208c` feat(daemon): cancel an authentication when its caller goes away — `CancelToken` checked per frame in auth/enroll/discard loops, per-request `NameOwnerChanged` watch on the sender, `cancelled` frozen string → PAM_IGNORE, audit `cancelled`, no limiter charge, suspend/`ReleaseCamera`/shutdown via the token
- `da3fcf0` fix(polkit): drop a dedicated bus connection when the dialog is cancelled
- `18b8ee8` fix(pam): one-shot loads models first, dies with its host, and cleans up — engine before camera, SIGTERM/SIGINT/SIGHUP → clean `Drop` (exit 2), `PR_SET_PDEATHSIG` in `pre_exec`, PAM child timeout TERM-then-KILL (≤500 ms)
- `8361d20` feat(daemon): end no-face attempts early and stop charging them — `recognition.no_face_timeout_secs` (clamped to `timeout_secs`, `0` disables, additive), no-face failures don't charge on daemon or one-shot path
- `8520ce0` feat(bench): camera-reopen microbench; container assertions for ADR 008 — `facelock bench camera-reopen` (open / first-frame / warmup split), doc numbers replaced by "measure it", ADR-tagged assertions in `test/run-integration-tests.sh` and `test/run-oneshot-tests.sh`
- `93364e4` fix(daemon): per-request cancel token; audit and classify every cancellation — review H1/H2/M1–M3: a fresh `CancelToken` per D-Bus call (departure watch bound to it alone; a denied or concurrent caller can no longer clear or trigger another request's cancellation), generation-counted `CurrentRequest` slot for suspend/`ReleaseCamera`, warm-discard cancellation returns the frozen `cancelled` string, pre-open cancellations audited, `Cancelled` rows added to both outcome tables
- `a25b331` fix(pam,cli,docs): oneshot child reaping and PDEATHSIG race; doc corrections
- `44cad3a` feat(daemon): opt-in camera hold after a successful attempt — `device.camera_release_after_success_secs` (default `0` = release immediately, unchanged behavior; `N` = keep warm N s after a success too, for repeated privileged actions with no auth caching, e.g. `sudo` with `timestamp_timeout=0`). Added on maintainer request as the knob ADR §3 had deferred; cancel/error still never hold
- `f33bd78` docs(adr): accept ADR 008 — Status → Accepted, verified as below — review L1/L2/L5/L6/L7: reap after SIGKILL, `getppid()` check after `prctl` closes the fork/exec race, `bench` root-gate docs made truthful, ADR §8 suspend wording matches the code, container scripts restore the original `device_id` and use a locale-independent success signal

## Verification

Ran on the full branch (Arch, host toolchain per `rust-toolchain.toml`) — all green:
- `cargo fmt --all -- --check`
- `cargo test --workspace` (14/14 test binaries; new tests: outcome×finish matrix over every `AuthOutcome`/`ErrorKind` variant, expire timing, discard count, cancel-at-frame-k, name-owner-changed sender matching, no-face timing, rate-limit charge/no-charge, engine-before-camera order, PAM TERM-then-KILL sequence)
- `cargo build -p pam-facelock` (isolated) + dependency-surface guard (57 crates, no nix/signal-hook)
- `just test-arch-pam` (camera-free Arch container PAM smoke tier): 22/22 passed
- `bash -n test/run-integration-tests.sh test/run-oneshot-tests.sh`
- CI on the final push (`a25b331`): Build/Test/Lint, Container PAM Smoke, cargo-audit, TPM tests — all green

Commits `11bcb59`..`18b8ee8` were additionally verified green in the cloud sandbox (Ubuntu) where they were authored.

## Camera container tiers (run locally by Ty on the BRIO, 8-commit branch)

- `just test-arch-integration`: **26/26 passed** — ADR 008 assertions with measured timings on the live daemon: success → camera released **3 ms** after the reply; failed attempt → warm hold expired **3206 ms** after the reply (`camera_release_secs=3`, 250 ms poll); client killed mid-`Authenticate` → departure seen at 2 ms, attempt **cancelled at 107 ms** (one frame), camera released 3 ms later.
- `just test-arch-oneshot`: **27/27 passed** — SIGTERM ends `facelock auth` with **exit 2 in 486 ms**, no process left; `facelock auth` gone **507 ms** after its PAM host was SIGKILLed (PDEATHSIG). Both signals landed during model load, where the token is first checked, so these are engine-load-bounded, not scan-loop latency.

## Host verification (Ty, branch installed via `just install` on the BRIO desktop)

- Default config: LED goes off immediately on a successful match (design: success releases at once). ✔
- `[device] camera_release_after_success_secs = 5`: LED stays on ~5 s after a successful match, then goes off; picked up by live reload without a daemon restart. ✔
- Note for others reading the config: `no_face_timeout_secs` is the empty-chair timer under `[recognition]`, not a hold.

## Still open (do not block merge; tracked in ADR §11)

- Quattro + omarchy PR #6863: password-first unlock → LED off with the unlock (verifies quickshell's `PamContext.abort()` drops the PAM helper's bus connection so the name-owner watch fires; if not, add the deferred explicit `Cancel()` D-Bus method)
- `sudo` Ctrl-C inside `pam_authenticate`: which path fires (cancelled vs. normal failure/warm) — either satisfies the design
- `sudo facelock bench camera-reopen` on the BRIO and a quirked IR camera — decides whether the 3 s default should drop to 2 s

## Review

Independent opus review of `3158f37..8520ce0` against ADR 008 (posted as a PR comment): 2 high, 3 medium, 7 low. H1/H2/M1–M3 and L1/L2/L5/L6/L7 are fixed in `93364e4` / `a25b331`; L3 (second-signal forced shutdown for a hand-run `facelock auth`) deliberately deferred; L4 resolved by `f33bd78` (ADR accepted) except §11 Q1, which awaits the bench numbers.

## Deviations from ADR 008

None known.

## Contract changes (docs/contracts.md updated in-commit)

- `device.camera_release_secs`: semantics narrowed to "after a failed attempt", default 5 → 3, `0` honored (was silently 5)
- `device.camera_release_after_success_secs`: new, additive, default 0 (opt-in hold after success)
- `recognition.no_face_timeout_secs`: new, additive, default 2
- audit `result` gains `cancelled`; frozen error string `cancelled` (PAM → PAM_IGNORE)
- no new D-Bus methods, no exit-code changes, no PAM defaults changed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **One-shot `facelock auth` no longer outlives its PAM host or lights the
+  camera during the model load** (ADR 008 §7). Three changes, no new exit
+  code and no change to the existing one: the ONNX engine now loads *before*
+  the camera opens, so the IR LED is lit for the scan rather than for the
+  model load as well; SIGTERM, SIGINT and SIGHUP end the scan through the same
+  cancel token, which lets `Drop` run STREAMOFF and turn the emitter off
+  before exiting 2 with a `cancelled` log line; and the PAM module sets
+  `PR_SET_PDEATHSIG = SIGTERM` on the child, so a killed PAM host (an aborted
+  locker helper, a killed `sudo`) takes the helper with it instead of leaving
+  it scanning, reparented to init, with the camera on. The PAM module's own
+  timeout now sends SIGTERM and waits up to 500 ms before SIGKILL; it used to
+  SIGKILL immediately, which skips `Drop` and can leave an XU-controlled IR
+  emitter lit.
 - **An authentication whose caller has gone away now ends within one frame**
   (ADR 008 §5). Nothing could shorten the scan loop before: when a screen
   locker aborted PAM because the password was typed first, or a `sudo` was
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   compares the target user's marker against the database and prints a
   diagnostic when they disagree (or when the marker is unreadable), pointing
   at `sudo facelock setup` to reconcile. `is-enrolled` itself is unchanged.
+- **Opt-in camera hold after a successful authentication**: a new
+  `device.camera_release_after_success_secs` (default **0**) keeps the camera
+  streaming for that many seconds after a success, the way
+  `camera_release_secs` already does after a failure. At its default — which is
+  the recommended value and what every install gets without touching the file —
+  nothing changes: a success ends the interaction, so the stream (and on IR
+  hardware the emitter LED) goes out with the reply. It exists for the one
+  shape that really does re-authenticate immediately: privileged actions
+  repeated with no authentication caching in front of them, `sudo` with a zero
+  `timestamp_timeout` or a polkit action without `auth_admin_keep`, where each
+  action is a fresh authentication that would otherwise pay a camera reopen.
+  Failed attempts keep using `camera_release_secs`; cancellations and errors
+  release the camera at once whatever both keys say. Additive with a serde
+  default: no existing config is rejected, and the daemon reads it per request,
+  so no restart is needed.
 - **`facelock bench camera-reopen`**: measures what it actually costs to go
   from a closed camera to the first frame an authentication can analyze, split
   into device open + format negotiation, `STREAMON` + first frame, warmup
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **The daemon holds the camera open only after a failed authentication**
+  (ADR 008). Previously every request — success included — left the V4L2 stream
+  live for `device.camera_release_secs`, which on IR hardware is a visible
+  emitter LED burning for five seconds after the screen had already unlocked.
+  Success, cancellation and every error class now release the camera as the
+  request returns; only a no-match or timeout keeps it warm, because that is
+  the one ending a retry plausibly follows. The default drops from **5 to 3
+  seconds**, `0` now means *never hold* instead of being silently substituted
+  with 5, and the release is polled every 250 ms against an absolute deadline
+  rather than once a second. A warm reuse discards the stale V4L2 buffers
+  before analyzing anything, so a fresh attempt can never match on the tail of
+  the previous one. Preview frames keep their own floor of
+  `max(camera_release_secs, 2s)` so a live preview never reopens per frame.
+  **No action required on upgrade**: the key is unchanged in name and type, the
+  shipped config template has it commented out, and the daemon re-reads it per
+  request.
 - **`facelock status` says "cannot determine" instead of guessing**: a section
   whose probe failed — an unreadable database, a config that did not parse —
   now reports exactly that, and a daemon that is unreachable can never render
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **An authentication at an empty chair ends early and costs no rate-limit
+  budget** (ADR 008 §3/§4). A new `recognition.no_face_timeout_secs` (default
+  **2**) ends an attempt once that many seconds have passed with no face
+  detected at all; `recognition.timeout_secs` still bounds the slower case a
+  timeout is actually for — a face was seen and has not matched yet. A laptop
+  opened in front of nobody therefore lights its IR emitter for 2 seconds
+  instead of 5. The key is additive with a serde default, is clamped to
+  `timeout_secs` rather than validated against it, and `0` disables the early
+  exit, so no existing `/etc/facelock/config.toml` needs to change. Separately,
+  and regardless of that timeout, a failed attempt in which the camera never
+  saw a face no longer calls the rate limiter on either the daemon or the
+  one-shot path: an empty chair is not a guess, and charging it let a locker
+  that starts face auth on every wake spend the user's whole 5-attempt budget
+  before they sat down — the real attempt then met a lockout. A face that
+  *was* seen and did not match is charged exactly as before. The early ending
+  reports the same outcome the full timeout reports, so no client, sentinel or
+  audit result gains a case.
 - **One-shot `facelock auth` no longer outlives its PAM host or lights the
   camera during the model load** (ADR 008 §7). Three changes, no new exit
   code and no change to the existing one: the ONNX engine now loads *before*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   compares the target user's marker against the database and prints a
   diagnostic when they disagree (or when the marker is unreadable), pointing
   at `sudo facelock setup` to reconcile. `is-enrolled` itself is unchanged.
+- **`facelock bench camera-reopen`**: measures what it actually costs to go
+  from a closed camera to the first frame an authentication can analyze, split
+  into device open + format negotiation, `STREAMON` + first frame, warmup
+  discard, and first usable frame, over `--iterations` cycles (default 5). This
+  is the number `device.camera_release_secs` trades LED-on time against —
+  holding the stream warm after a failed attempt buys a retry exactly this
+  much (ADR 008) — and it had never been measured: the docs asserted "~400 ms"
+  and "~600 ms cold" from nobody's hardware in particular. Those figures are
+  gone from the docs in favor of pointing at this subcommand, since the answer
+  is a property of the camera and its driver, not of facelock. Needs no
+  enrolled face and loads no models.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **An authentication whose caller has gone away now ends within one frame**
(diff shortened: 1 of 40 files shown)


tyvsmith:
## Independent review of `3158f37..8520ce0` against ADR 008 (opus reviewer, read-only)

Fixes for H1/H2/M1–M3 and L1/L2/L5/L6/L7 are being pushed as follow-up commits; L3/L4 intentionally deferred (L4: ADR stays Proposed until merge).

## Verdict

The branch implements ADR 008 closely and the docs/CHANGELOG work is unusually thorough. `cargo build --workspace`, `cargo test --workspace`, and `cargo clippy --workspace --all-targets -- -D warnings` are all clean; `bash -n` passes on both container scripts; no manifest or `Cargo.lock` changes at all, so no dependency-ceiling movement.

One structural defect undercuts the branch's headline feature, plus a handful of smaller drifts.

---

## Findings

### H1 — The cancel token is daemon-global, not per-request, and `arm()` runs before authorization

`crates/facelock-daemon/src/server.rs:534` holds one `cancel: CancelToken` for the whole service. The `#[interface]` glue arms it at `server.rs:1046` (`authenticate`), `:1074` (`test_authenticate`), `:1098` (`enroll`) — **before** `authenticate_as`/`enroll_as`, which is where `authorize_method` (`:695`, `:766`) and `capture_slot.try_acquire` (`:697`) live.

zbus 5.16 dispatches each method call in its own spawned task — `spawn_tasks_for_methods` defaults to `true` (`~/.cargo/registry/.../zbus-5.16.0/src/object_server/interface/mod.rs:46`) and the daemon does not set `spawn = false` — so concurrent calls are real, which is exactly what `CaptureSlot` exists for.

Failure scenario: hyprlock's `Authenticate` is scanning; the user types the password first and the shell aborts PAM; the departure watch sets the token. Before the loop's next iteration reads it, any second D-Bus call — a retry, a polkit dialog, or a non-root `Enroll` that is about to be *denied* — runs `self.cancel.arm()` and clears the flag. The scan runs to `timeout_secs` with the IR emitter lit: the precise bug ADR 008 exists to fix, reintroduced by a request that never got past authorization.

The mirror case is worse: the rejected caller also subscribes a `NameOwnerChanged` watch against the **shared** token. If that caller exits inside the window before `_watch` drops (`server.rs:153-159`), it cancels the *victim's* in-flight request. Both directions are cross-request interference available to any `facelock`-group member, and both are invisible to the tests — `a_cancelled_authenticate_charges_no_rate_limit` (`crates/facelock-daemon/tests/server_authz.rs:558`) calls `authenticate_as` directly and never exercises the glue.

**Fix:** mint a `CancelToken` per request in the glue and hand it to the handler for the request's duration (the lease adopts it in `acquire`, releases in `finish`). If that is too large, move `arm()` and the watch subscription to *after* `authorize_method` and `capture_slot.try_acquire()` both succeed — the capture slot is already the per-request mutual exclusion this needs.

### H2 — A cancelled request leaves the token latched, wedging every later preview frame

`CameraLease::finish` (`crates/facelock-daemon/src/handler.rs:409-430`) never re-arms the token; only `release()` does (`:465`). `preview_frame_as` / `preview_detect_frame_as` do not arm either. So after any cancelled authentication the flag stays `true`, and `acquire` (`:358`) short-circuits every subsequent `PreviewFrame` with `Err("cancelled")`.

Scenario: a locker aborts PAM → auth ends `Cancelled` → an admin runs `facelock preview` → every frame errors until something arms the token (the preview's own `ReleaseCamera` on exit, or the next auth). The same latch persists when the suspend path's `try_lock` times out (`server.rs:1461-1467`).

The team already identified this shape — the test comment at `handler.rs:1249-1254` says *"which is how a `ReleaseCamera` used to be able to wedge every later preview frame"* — but `finish(Cancelled)` reintroduces it on a different path.

**Fix:** have `finish` call `self.cancel.arm()` on the close path, for the same reason `release()` does. Per-request tokens (H1) make it moot.

### M1 — Cancellation during the *warm* stale discard reports `Error`, not `Cancelled`

`handler.rs:368-378`: `discard_frames` returns `Err(CANCELLED_MESSAGE)` on a set token, and the warm branch wraps it as `format!("capture error: {e}")` → `"capture error: cancelled"`. The cold-warmup branch gets this right (`:391-398` returns the bare `CANCELLED_MESSAGE`).

ADR §8 row 2 specifies "Cancel token during warmup/stale discard → abort acquire → `Cancelled`, closed". The camera does close, so the §8 invariant holds, but no `cancelled` audit row is written and PAM's exact-match on `cancelled` misses, landing on the generic `"error: internal"` branch instead.

**Fix:** check `self.cancel.is_cancelled()` in the warm branch before wrapping and return `CANCELLED_MESSAGE` verbatim.

### M2 — A cancellation before the camera opens is never audited

`handle_authenticate` (`handler.rs:921-929`) returns `DaemonResponse::Error` directly when `acquire` reports cancelled. `auth::cancelled()` (`auth.rs:930-960`), the only writer of the `result: "cancelled"` audit row, is reached only from inside the scan loop. `docs/contracts.md` advertises `cancelled` as covering "the caller's bus connection went away, the system suspended, `ReleaseCamera` arrived" — the pre-open case silently writes nothing.

### M3 — The two "every outcome" tables omit the variant this branch adds

`handler.rs:1088-1106` covers `AuthResult(true/false)`, `Suppressed`, and every `ErrorKind`, but not `AuthOutcome::Cancelled`. `handler.rs:1109-1124` covers `Enrolled` and `Error`, but not `EnrollOutcome::Cancelled`. The wildcard-free `From` impls still force compile-time classification, but the assertions ADR §9 asks for don't actually cover the new rows. Two added asserts.

### Lower severity

- **L1** `terminate_oneshot_child` (`crates/pam-facelock/src/lib.rs:825-855`) leaves a zombie on the SIGKILL path: one non-blocking `try_wait()` immediately follows the kill, and `Child::drop` detaches. Pre-existing behavior, not a regression, but one `waitpid` from fixed.
- **L2** PDEATHSIG has the classic fork/exec race — if the host dies between `fork` and the `prctl`, the child never gets SIGTERM, and the 500 ms kill sequence can't help because the host is already gone. Standard mitigation is `if getppid() != expected { raise(SIGTERM) }` right after the prctl; `getppid` is async-signal-safe so it fits the constraint. The `pre_exec` closure itself is correct: one `prctl` syscall, no allocation, no locks.
- **L3** `signal_hook::flag::register` makes a hand-run `facelock auth` ignore Ctrl-C if the scan wedges in the driver. The PAM path is covered by SIGTERM→500 ms→SIGKILL, and `CAPTURE_TIMEOUT` (2 s, `crates/facelock-camera/src/capture.rs:15`) bounds it in practice. `register_conditional_shutdown` for the second signal would close it.
- **L4** `docs/adr/008-camera-lifecycle.md:5` is still `Status: Proposed` while the branch ships all six commits, and §11 Q1 (3 s vs 2 s) is unresolved despite the bench existing to answer it.
- **L5** `docs/cli.md:215` and `book/src/cli-reference.md:428` claim `camera-reopen` "is still root, like every `bench` subcommand". There is no root gate on `bench` in the CLI — the others fail only because they read root-owned DB/model files (`crates/facelock-cli/tests/cli_smoke.rs:85`). `camera-reopen` reads neither, so a `video`-group user can run it. Reword or add the gate.
- **L6** ADR §8 promises suspend will "drop the camera anyway" after 1 s; `server.rs:1461-1467` only warns and returns (it cannot take the mutex). The code is the honest choice; the ADR text overpromises.
- **L7** Both container scripts forge `device_id` and restore with `device_id=NULL` rather than the original value. Harmless because `bind_legacy_templates` defaults true (`crates/facelock-core/src/config.rs:355`), but it isn't a restore. Also `adr008_success_releases_camera` greps for `Matched`, which goes through `translate()` (`crates/facelock-cli/src/message/face.rs:178`) — locale-dependent, though it fails *safe* to SKIP.

---


🤖 Posted by Claude Code

tyvsmith:
Resolution of the review above: H1/H2/M1–M3 fixed in 93364e4 (per-request cancel token, generation-counted current-request slot, warm-discard cancel returns the frozen `cancelled` string, pre-open cancellations audited, `Cancelled` rows in both outcome tables); L1/L2/L5/L6/L7 fixed in a25b331; L4 closed by f33bd78 (ADR accepted; §11 Q1 still awaits `bench camera-reopen` numbers). L3 intentionally deferred. Live-camera container tiers subsequently run on the BRIO: integration 26/26, oneshot 27/27, ADR timing assertions measured (success release 3 ms, hold expiry 3206 ms @3 s, departed-caller cancel 107 ms, one-shot SIGTERM exit 486 ms, PDEATHSIG 507 ms).

🤖 Posted by Claude Code

PR #245, opened 2026-08-20
fix(rpm): harden onnx runtime loading

tyvsmith:
- enforce trusted privileged ONNX Runtime loading because package and daemon contexts must never honor attacker-selected libraries
- separate checksum-pinned bundled direct RPMs from Fedora system-runtime COPR builds
- keep Rawhide best-effort, nongating, and excluded from release publication
- validate with just check, just test-rpm-pkg, just test-copr, and the RPM policy gate


tyvsmith:
- include the exact commit in the merged native stack through #254
- close this duplicate PR after landing

PR #298, opened 2026-08-25
fix(arch): declare onnxruntime in facelock-git and pin the source digest

### LINKED ISSUE #283: dist/PKGBUILD skips the sha256 of a fixed release tarball
`dist/PKGBUILD` fetches a fixed release tarball over https and declares
`sha256sums=('SKIP')`, so `makepkg` verifies nothing about what it downloaded.

## Where

- `dist/PKGBUILD:19` — `source=("$pkgname-$_tag.tar.gz::$url/archive/v$_tag.tar.gz")`
- `dist/PKGBUILD:20` — `sha256sums=('SKIP')`

## Expected vs actual

Expected: a fixed, immutable artifact is pinned by digest, matching how the rest
of the project pins its inputs — GitHub Actions by commit SHA, container images by
digest, the ONNX Runtime bundle and the models by checksum.

Actual: the AUR `facelock` package accepts whatever that URL returns.

## Scope

`dist/PKGBUILD` only. `dist/PKGBUILD-bin:26` already carries real hashes.
`dist/PKGBUILD-git:17` uses `SKIP` correctly — a `git+` source cannot be hashed.

Surfaced by the new Arch end-to-end lane, which parses and exercises the field but
cannot assert anything about a skipped sum.

### LINKED ISSUE #284: dist/PKGBUILD-git omits the onnxruntime dependency the other two recipes declare
`dist/PKGBUILD-git` does not list `onnxruntime` in `depends` or `makedepends`,
while `dist/PKGBUILD` and `dist/PKGBUILD-bin` both list it in `depends`.

## Where

- `dist/PKGBUILD-git:9` — `depends=('glibc' 'dbus' 'pam' 'gcc-libs' 'tpm2-tss' 'libxkbcommon')`
- `dist/PKGBUILD-git:10` — `makedepends=('rust' 'cargo' 'clang' 'wayland' 'libxkbcommon' 'git')`
- `dist/PKGBUILD:10` and `dist/PKGBUILD-bin:11` both include `onnxruntime`

## Expected vs actual

Expected: installing `facelock-git` pulls in the ONNX Runtime the daemon loads at
startup, as the other two recipes do.

Actual: it does not, so a machine with no other source of `libonnxruntime` would
install the package and fail at daemon start.

## Scope

`facelock-git` is the recipe Omarchy's installer uses
(`omarchy-pkg-aur-add facelock-git`), so this is the path most affected.

Not verified whether the omission is deliberate — e.g. if the git build resolves
the runtime differently. If it is deliberate, the reason belongs in a comment,
since the three recipes otherwise agree.

tyvsmith:
## Summary

- add `onnxruntime` to `dist/PKGBUILD-git` depends and optdepends, matching the other two recipes
- replace `sha256sums=('SKIP')` in `dist/PKGBUILD` with a fail-closed `__SRC_SHA256__` placeholder, finalized by `publish-aur.sh` at publish
- refuse a malformed or empty-input checksum in the publisher; fetch the release tarball with `curl -f`
- teach the Arch e2e lane to reject `SKIP`, prove a wrong digest fails `--verifysource`, then verify the staged tarball's real digest
- guard both contracts in the matrix checker, with contract-test mutations proving the gates bite
- copy the four files the matrix checker reads since 6e00813 into the contract fixture; their absence failed `test/release-version-contract.sh` on main

## Why

`ort` builds with `load-dynamic`, so libonnxruntime is dlopened at first authentication rather than linked at build time: `facelock-git`, the recipe Omarchy's installer pulls, installed cleanly on a machine with no ONNX Runtime and failed at first login. `dist/PKGBUILD` was the one fetched artifact in the project trusted without a digest, and the tagged tree cannot carry its own tarball's sum, so the pin follows the `PKGBUILD-bin` convention: placeholder in-repo, real digest substituted at publish.

## Validation

- `just check`
- `just test-arch-pkg`
- `bash test/release-version-contract.sh`
- `python3 test/check-release-matrix.py`
- `makepkg --verifysource` against a corrupted tarball, with the old `SKIP` recipe and with a pinned digest

Fixes #283
Fixes #284

## Summary by CodeRabbit

- **Bug Fixes**
  - Strengthened Arch and AUR package integrity checks with required, validated SHA-256 checksums.
  - Release publishing now fails clearly when downloads fail or checksum data is invalid.
  - Arch package variants consistently include the required runtime dependency.

- **Documentation**
  - Updated packaging, release, and testing guidance for checksum verification and publish-time substitution.
  - Clarified remaining package-testing limitations, including source URL and system startup coverage.

### DIFF
--- a/.claude/skills/packaging-test/SKILL.md
+++ b/.claude/skills/packaging-test/SKILL.md
@@ -67,17 +67,18 @@ hook that cleans PAM up on removal. It also resolves every dependency name all
 three PKGBUILDs declare, including `PKGBUILD-git`, which is what Omarchy's
 `omarchy-pkg-aur-add facelock-git` pulls.
 
-Three things it does not do. It does not boot systemd, so unit runtime
-behaviour stays with the deb and rpm gates. It checks no source digest:
-`sha256sums` is `SKIP` (#283), the staged tarball is a repack of the working
-tree, and the `source=` URL is never fetched, so a wrong URL passes. And it
-compiles the workspace twice, release for `build()` and debug for `check()`, so
-it is the slowest recipe here. Reach for `test-arch-pam` while iterating and run
-this before pushing a packaging change.
+It also exercises integrity the way the shipped recipe gets it: `dist/PKGBUILD`
+carries a fail-closed `__SRC_SHA256__` placeholder that `publish-aur.sh`
+finalizes at release time (#283), so the lane refuses a recipe declaring
+`SKIP`, proves a wrong digest is rejected, then substitutes the staged
+tarball's real digest and lets `makepkg` verify it.
 
-When #283 replaces `SKIP` with a real digest, this lane breaks: a staged tarball
-cannot match a published sum. Move the staged build to `--skipchecksums` and
-assert the real digest separately.
+Two things it does not do. It does not boot systemd, so unit runtime
+behaviour stays with the deb and rpm gates. And it compiles the workspace
+twice, release for `build()` and debug for `check()`, so it is the slowest
+recipe here. Reach for `test-arch-pam` while iterating and run this before
+pushing a packaging change. The `source=` URL is still never fetched, so a
+wrong URL passes.
 
 ## Camera-gated tests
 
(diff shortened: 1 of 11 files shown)


tyvsmith:
@codex review this.

PR #299, opened 2026-08-25
fix(preview): resolve the wayland socket from the invoking uid

### LINKED ISSUE #297: facelock preview never opens the documented Wayland window
`facelock preview` is documented as opening a Wayland graphical window, but
always falls back to text-only. The root re-exec drops the two environment
variables the Wayland connection needs, so neither documented invocation can
reach the compositor.

## Reproduction

Under a Wayland session, either form:

```
sudo facelock preview
facelock preview          # prompts, re-execs itself under sudo
```

Both print:

```
WARN Wayland preview failed: failed to connect to Wayland display
Wayland preview unavailable: failed to connect to Wayland display
Falling back to text-only mode.
```

## Expected vs actual

Expected: `docs/cli.md:353` describes `facelock preview` as "Wayland graphical
window", so a Wayland session should get one.

Actual: text-only, every time.

## Suspected cause

`crates/facelock-cli/src/commands/preview/wayland_preview.rs:46` calls
`Connection::connect_to_env()`, which reads `WAYLAND_DISPLAY` and
`XDG_RUNTIME_DIR`. `crates/facelock-cli/src/ipc_client.rs:44-49` re-execs
`sudo <args>` with no environment preservation, and `sudo`'s `env_reset`
strips both. Running `sudo facelock preview` directly loses them the same way.

Root can open the session's Wayland socket; it is not told where to find it.

## Scope

Preview only. The text-only path works and is what the daemon's
`PreviewDetectFrame` drives, so nothing is broken beyond the graphical mode
being unreachable. The command is root-only, so this cannot be fixed by simply
dropping the root requirement.

## Possible approach

Preserve exactly `WAYLAND_DISPLAY` and `XDG_RUNTIME_DIR` across the re-exec —
narrowly, not a general env passthrough, given the re-exec targets root.
Alternatively document the fallback as the expected behaviour and stop
advertising a window.

tyvsmith:
## Summary

- resolve the preview's Wayland socket from the invoking uid's `/run/user/<uid>` instead of `Connection::connect_to_env()`
- take the uid from `SUDO_UID`/`DOAS_USER`, falling back to the current uid for a root login
- authenticate the connected peer: `SO_PEERCRED` must show a process running as the invoking uid before any protocol is spoken; the lstat symlink/ownership checks stay as a fast filter only, since the user-writable runtime dir makes any pathname check raceable
- honour `WAYLAND_DISPLAY` only as a bare socket name inside that directory; ignore values carrying a path separator
- carry `WAYLAND_DISPLAY` across the preview's sudo re-exec (`--preserve-env=WAYLAND_DISPLAY`), so a session with several compositors previews on the one that ran the command; every other command's re-exec is byte-identical to before
- document the resolution in `docs/cli.md`

## Why

Both documented invocations run the preview as root, and sudo's `env_reset` strips `WAYLAND_DISPLAY` and `XDG_RUNTIME_DIR`, so `connect_to_env()` never reached the compositor and every preview fell back to text-only. Preserving the runtime dir into the root re-exec was rejected: a hostile `XDG_RUNTIME_DIR` points root's socket connect at an arbitrary path, the shape the PAM oneshot spawn refuses. Identity-based resolution never reads it, and peer-credential authentication bounds what remains: root only ever speaks Wayland to a process running as the invoking uid, and the worst a hostile value or path swap achieves is a bare connect-and-close or the text-only fallback.

## Validation

- `just check`
- `cargo test --workspace`
- live on a Wayland session: `sudo facelock preview` (scan path) and `sudo --preserve-env=WAYLAND_DISPLAY facelock preview` (named path) both open the layer-shell window and exit cleanly on SIGINT
- live negative: a planted `/run/user/1000/wayland-9` symlink to a root-owned socket, selected by a hostile display name, is refused and degrades to text-only

Fixes #297

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Wayland preview support when running with elevated privileges.
  * Preview now connects to the invoking user’s compositor and validates socket ownership for safer, more reliable startup.
  * Preserves the selected Wayland display during privilege escalation and supports automatic compositor discovery.

* **Documentation**
  * Documented root-invoked Wayland behavior, display selection, and text-only fallback handling.

### DIFF
--- a/crates/facelock-cli/Cargo.toml
+++ b/crates/facelock-cli/Cargo.toml
@@ -43,7 +43,7 @@ dialoguer = "0.12"
 signal-hook = "0.4"
 thiserror = { workspace = true }
 libc = "0.2"
-nix = { version = "0.31", features = ["user"] }
+nix = { version = "0.31", features = ["user", "socket"] }
 facelock-tpm = { path = "../facelock-tpm" }
 zbus = { version = "5", default-features = false, features = ["tokio", "blocking-api"] }
 tss-esapi = { workspace = true, optional = true }
--- a/crates/facelock-cli/src/commands/preview/mod.rs
+++ b/crates/facelock-cli/src/commands/preview/mod.rs
@@ -3,6 +3,8 @@ mod render;
 mod text_only;
 #[cfg(feature = "wayland")]
 mod wayland_preview;
+#[cfg(feature = "wayland")]
+mod wayland_socket;
 
 use facelock_core::Config;
 
--- a/crates/facelock-cli/src/commands/preview/wayland_preview.rs
+++ b/crates/facelock-cli/src/commands/preview/wayland_preview.rs
@@ -43,7 +43,10 @@ pub fn run(user: &str) -> anyhow::Result<()> {
     signal_hook::flag::register(signal_hook::consts::SIGINT, std::sync::Arc::clone(&stop))
         .context("failed to register SIGINT handler")?;
 
-    let conn = Connection::connect_to_env().context("failed to connect to Wayland display")?;
+    // Resolved from the invoking user's identity, not `connect_to_env()`:
+    // the preview runs as root and sudo's `env_reset` had stripped the
+    // variables that function needs, so it never connected (issue #297).
+    let conn = super::wayland_socket::connect()?;
 
     let (globals, mut event_queue) =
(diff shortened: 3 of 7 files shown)


tyvsmith:
@codex review

PR #300, opened 2026-08-25
fix(face): resolve symlinked ORT trust roots before validation

### LINKED ISSUE #295: ONNX Runtime candidates under a symlinked /usr/lib64 are rejected unresolved
`validate_directory` rejects any trusted ONNX Runtime directory that is a
symlink, without resolving it. On a merged-`/usr` distribution where
`/usr/lib64` is a symlink to `lib`, the `/usr/lib64` candidate is refused even
though it resolves to a root-owned, non-writable real directory.

## Reproduction

On CachyOS (`/usr/lib64 -> lib`, root-owned, `drwxr-xr-x`), with a system
ONNX Runtime at `/usr/lib/libonnxruntime.so.1.28.0`:

```
sudo facelock auth --user <user>
WARN Rejected ONNX Runtime candidate /usr/lib64/libonnxruntime.so: trusted directory /usr/lib64 is a symbolic link
```

## Expected vs actual

Expected: a symlinked trusted directory is resolved and the *resolved* target
validated for ownership and mode, since that is the property the check exists to
enforce.

Actual: `crates/facelock-face/src/provider.rs:398-402` returns an error on
`file_type().is_symlink()` before any resolution, so the candidate is dropped.

## Scope

Benign where another candidate resolves — here `/usr/lib` succeeded and
authentication completed. It becomes a hard failure on a distribution where the
symlinked path is the *only* location the runtime is installed to.

Merged-`/usr` layouts with `/usr/lib64 -> lib` are the common case on Arch and
derivatives; Fedora ships a real `/usr/lib64` directory and is unaffected.

Note the adjacent link-validation helper at `provider.rs:481-493` already
verifies a symlink's owner and hard-link count, so the building blocks for
resolving safely exist.

tyvsmith:
## Summary

- resolve a symlinked ONNX Runtime trust root by path before the strict checks, under the rules already applied to SONAME links: root-owned single-link symlinks, confined relative targets, every traversed directory root-owned and not group- or world-writable
- reject absolute and escaping link targets even when they name an approved location
- run the unchanged directory, owner, and mode checks on the resolved real directory; the `O_NOFOLLOW` open and descriptor-held validation stay the load-bearing gate
- cover accept, writable target, escaping target, dangling link, loop, and wrong owner in unit tests, plus a merged-/usr host regression test
- describe the resolution step in `docs/security.md` and `docs/contracts.md`

## Why

On merged-/usr distributions `/usr/lib64` is a root-owned symlink to `lib`, and the loader rejected the candidate without resolving it. That is benign where another candidate resolves, and a hard failure where the symlinked path is the only location the runtime is installed to. Only root can retarget a root-owned symlink beneath root-owned, non-writable ancestors, so the resolved directory carries the same trust as a real one.

## Validation

- `just check`
- `cargo test --workspace`
- privileged loader gate against a live merged-/usr `/usr/lib64 -> lib` layout

Fixes #295

## Summary by CodeRabbit

- **Security**
  - Improved privileged ONNX Runtime loading for trusted, symlinked system library locations.
  - Supports confined merged-`/usr` layouts while rejecting unsafe links, including absolute, escaping, dangling, untrusted, or writable paths.
  - Maintains strict ownership, directory, and link-validation checks during library loading.

- **Documentation**
  - Updated security and contract documentation to describe trusted-root symlink handling and validation requirements.

- **Tests**
  - Added coverage for approved symlink layouts, invalid targets, and host filesystem configurations.

### DIFF
--- a/crates/facelock-face/src/provider.rs
+++ b/crates/facelock-face/src/provider.rs
@@ -599,6 +599,99 @@ fn confined_components(path: &Path) -> std::result::Result<VecDeque<OsString>, S
         .collect())
 }
 
+/// Resolve a trusted root below `base` by path, following only approved-owner
+/// directory symlinks with confined relative targets.
+///
+/// Merged-/usr distributions ship `/usr/lib64` as a symlink to `lib`, so an
+/// approved root can be a link whose resolution is a root-owned real
+/// directory. A symlink's own mode carries no trust on Linux — it is always
+/// 0o777 and never consulted — so the facts that make a link as trustworthy
+/// as a real directory are the ones `validate_symlink_owner` already demands
+/// of SONAME links: the link is owned by the approved uid, has a single name,
+/// and every directory that could rename or retarget it is owned by the
+/// approved uid and not group- or world-writable. Under those facts only the
+/// approved uid chooses where the link points, exactly as only the approved
+/// uid chooses a real directory's entries.
+///
+/// Link targets must be confined relative paths — the same rule the
+/// descriptor walker applies beneath an opened root — so resolution only
+/// descends from `base` through components this walk has validated and can
+/// never leave the approved ancestry through `..` or an absolute target. An
+/// absolute target stays rejected even when it names an approved location;
+/// fail-closed beats a second list of approved spellings.
+///
+/// This is a by-path pre-check, like `validate_root_ancestors`. The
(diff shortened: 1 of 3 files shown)


tyvsmith:
@codex review

Issue #189, opened 2026-08-17
cli_smoke C6 rows are skipped in CI and depend on a config file locally


tyvsmith:
The C6 ordering rows in `crates/facelock-cli/tests/cli_smoke.rs` ("root check runs before any output") never execute in CI, and locally most of them depend on `/etc/facelock/config.toml` existing.

## Reproduction

1. `assert_refuses_before_output` returns early when `Uid::effective().is_root()`; the `build-and-test` job in `.github/workflows/ci.yml` runs in `archlinux:base-devel` as root, so all 16 rows are skipped there.
2. Locally: `FACELOCK_CONFIG=/nonexistent/nope.toml cargo test -p facelock-cli --test cli_smoke` → 8 rows fail (`audit`, `bench`, `clear`, `devices`, `list`, `preview`, `remove`, `test_cmd`) with "no config file at …" instead of "Root required": those commands hit `ConfigLoad::require()` before their own root check.

## Expected vs actual

The rows should gate C6 on every push. Today they pass vacuously in CI and are environment-dependent locally.

## Scope

Test infrastructure only; the C6 behaviour itself was verified by hand for the moved commands in #185. Needs a non-root step (or `su` to a test user) and a config fixture, and possibly a decision on whether the config load should come after the root check for those eight commands.

PR #301, opened 2026-08-25
test(cli): pin enroll's root gate and stop skipping under root

### LINKED ISSUE #288: enroll has no C6 smoke row and lost its in-command root check
`enroll` is the one gated command with no C6 ordering row in
`crates/facelock-cli/tests/cli_smoke.rs`, and it also lost the in-command root
check that previously backed up the central gate.

## Where

- `crates/facelock-cli/tests/cli_smoke.rs` — 18 C6 rows cover every other gated
  spelling; `enroll` has none
- `crates/facelock-cli/src/commands/enroll.rs` — the `require_root` calls that were
  at `:21` and `:37` before the C6 gate landed are gone, replaced by the
  `require_root_for` arm in `main.rs`

## Expected vs actual

Expected: every command the central gate covers has a row proving it refuses before
output, matching the other 18.

Actual: `enroll`'s ordering is enforced only by the gate's exhaustive match, with no
test that would fail if the arm were removed or mis-classified.

## Scope

`enroll::run` is also called internally from `setup.rs:2166,2186,2541`. Those paths
are reachable only through `run_with_plan`'s precheck (`setup.rs:397`, guarded by
`needs_root_precheck` = `plan.base.is_some()`), so they are gated today. The residual
is that a future setup path with `plan.base == None` calling `enroll::run` would skip
the root check silently, where previously the in-command check would have caught it.

No current path does this. Found during adversarial review of #264, which was
otherwise sound on every axis.

### LINKED ISSUE #303: pty escalation-class tests skip under the root CI runner
The two pty-based escalation-class tests return early when the test runner is
root, so they never execute in CI. They are the only tests that distinguish
`require_root` from `require_root_scripted`.

## Where

- `crates/facelock-cli/tests/cli_smoke.rs` — `daemon_run_never_prompts_with_a_tty_attached` (added by #262)
- `crates/facelock-cli/tests/cli_smoke.rs` — `enroll_offers_the_sudo_re_exec_with_a_tty_attached` (added by #288)

## Expected vs actual

Expected: the escalation class of every gated command is pinned on every push,
the way the C6 ordering rows now are.

Actual: both jobs in `.github/workflows/ci.yml` run as root in an
`archlinux:base-devel` container, so both tests skip. They pass vacuously
in CI and only really run on a developer's non-root machine.

## Scope

This is #189 in miniature. #272 fixed the same class of hole for
`assert_refuses_before_output` by dropping the spawned child to uid 65534
instead of skipping. The equivalent fix should work here: an inherited pty
file descriptor survives a `setuid` in the child, because the permission check
happened at `openpty` time in the parent.

Not done as part of #288 because the root path could not be exercised without
sudo, and an unverified root path risks hanging CI on the tests' 30s deadline —
an honest skip was judged better than an unproven change. Fixing it needs a run
as root to confirm the pty survives the uid drop and the deadline still holds.

tyvsmith:
## Summary

- add the C6 ordering row for `enroll`, the one gated spelling that had none
- point that row's `--config` at a path that does not exist, so only the central gate can answer first
- add a pty row pinning `enroll` to the interactive escalation class, the shape #262 used for `daemon run`
- re-check root on entry to `enroll::run`, hard-error class, for the `setup` callers that never reach the gate
- drop the spawned child to uid 65534 in both pty rows instead of returning early under a root runner
- re-execute the test binary at itself for the backstop test, which calls `run` in-process and had no child to drop

## Why

`enroll`'s ordering rested entirely on `require_root_for`'s exhaustive match, so
neutering the arm failed nothing. It also lost its in-command check when the
checks moved into the central gate, and `setup` still calls `enroll::run`
directly, behind a precheck conditional on a base flow running. Nothing reaches
it ungated today, but a future standalone plan that enrolled would, silently.

Every test added here would then have skipped on the one configuration CI runs.
Both CI test jobs are root in a container, and a row that returns early there
asserts nothing while reporting a pass, which is the defect #189 already fixed
once for the stdin-closed rows.

The two halves interact, which is why the rows pin `--config`: with a readable
config on the machine, the backstop answers "Root required" for a gate arm that
is already gone, one parse too late.

## Validation

- `just check`
- the smoke suite and the lib tests as a non-root user, and again as uid 0 in a user namespace where 65534 is mapped, which is what a root CI runner sees
- mutation checks under both uids, against the rows meant to catch them: neutering the gate arm, downgrading it to `require_root_scripted`, dropping the in-command check, and renaming the re-executed test so its filter matches nothing

Fixes #288
Fixes #303

## Summary by CodeRabbit

- **Bug Fixes**
  - Enrollment now consistently requires administrator privileges before accessing configuration or enrollment data.
  - Non-administrator attempts are rejected early, including enrollment paths launched through setup.
  - Interactive terminal sessions can offer sudo re-execution, while declining the prompt safely cancels enrollment.
  - Enrollment refuses to prompt when run non-interactively, such as through the daemon.
  - Improved handling of enrollment commands in automated and terminal-based environments.

- **Documentation**
  - Clarified administrator requirements and privilege-check behavior for enrollment commands.

### DIFF
--- a/crates/facelock-cli/src/commands/enroll.rs
+++ b/crates/facelock-cli/src/commands/enroll.rs
@@ -184,22 +184,113 @@ mod tests {
         config
     }
 
+    /// The uid/gid a root test runner drops the child to, matching
+    /// `tests/cli_smoke.rs`. `setuid`/`setgid` do not consult `/etc/passwd`,
+    /// so the account need not exist inside a CI container.
+    const NOBODY_UID: u32 = 65534;
+    const NOBODY_GID: u32 = 65534;
+
+    /// Set in the re-executed copy of this test binary, to tell it that it is
+    /// the child and should run the assertion rather than spawn again.
+    const BACKSTOP_CHILD_VAR: &str = "FACELOCK_TEST_ENROLL_BACKSTOP_CHILD";
+
+    /// The line the child prints once the refusal has held, with the uid it
+    /// held under. libtest exits 0 when a filter matches nothing, so a
+    /// successful child status is not on its own evidence that anything ran.
+    const BACKSTOP_WITNESS: &str = "enroll backstop refused as uid";
+
     /// Issue #288: the backstop at the top of [`run`] is the whole defense
     /// for callers that bypass `main`'s gate, and `setup` is one — it calls
     /// `run` directly, behind a precheck that is conditional on
     /// `plan.base.is_some()`. Nothing about the gate's exhaustive match
     /// would notice a future standalone plan enrolling unprivileged.
     ///
-    /// The store path is a temp dir that must survive untouched: the refusal
-    /// has to land ahead of every read `run` would otherwise do.
+    /// `run` is called in-process, so unlike the rows in `tests/cli_smoke.rs`
+    /// there is no child to drop privileges on — and a test that returned
+    /// early under a root runner would assert nothing in CI, where both test
+    /// jobs are root in a container (issues #189, #303). So this re-executes
+    /// the test binary at itself and drops *that* child, which puts the
+    /// assertion back under the same rule every spawning row follows.
+    ///
+    /// The re-exec is unconditional. A branch that only spawned under root
+    /// would leave the path CI takes as the one path no developer ever runs,
+    /// which is the shape of the bug being fixed.
     #[test]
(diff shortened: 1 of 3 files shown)


tyvsmith:
@codex review

PR #302, opened 2026-08-25
ci(packaging): gate every package lane in ci

### LINKED ISSUE #229: test(deb): exercise exact artifact install upgrade remove and purge lifecycles
## Why

Current Debian package tests preinstall dependencies, tolerate failed installation, mask tmpfiles behavior, and do not execute package transactions in a clean booted environment. Exact candidate artifacts need native install, upgrade, remove, and purge gates.

## Dependencies

- block on Tracks D through I
- block served-repository lanes on the named #236 staging APT endpoint

## Scope

- build suite-native candidate artifacts before booting the test environment
- install bind-mounted immutable artifacts at runtime through APT
- remove preinstalled runtime dependencies and `dpkg -i` repair masking
- provision checksum-pinned model fixtures without hiding installed ownership
- cover Trixie and Resolute variants
- cover active and inactive daemon, shared and direct PAM, arbitrary service, remove, reinstall, and purge paths

## Acceptance

- verify ownership, dependency closure, multiarch PAM, ORT initialization, tmpfiles, D-Bus, and systemd state
- verify correct and incorrect password behavior after every biometric failure
- verify remove preservation and purge deletion contracts
- fail on every skipped assertion or unexpected package-manager repair
- bind exact artifact and served-repository digests to the candidate commit

tyvsmith:
## Summary

- add `.github/workflows/packaging.yml`: both Debian suite gates, Fedora 43/44/45 at their declared lifecycle depths, the Arch package built from `dist/PKGBUILD`, and the native version-ordering matrix
- filter pull requests through a `changes` job over a merge-base diff in plain bash, so every lane reports a real `skipped` conclusion instead of stranding a required check
- run the same matrix unfiltered nightly at 07:00 UTC, and on `workflow_dispatch` for the release gate
- factor `container-pam-test`'s model download into `.github/actions/fetch-models`, checksum-verified against `models/manifest.toml` and cached on its hash
- gate `just release-preflight` on a green `packaging.yml` run at HEAD, or on the record `just test-packaging-matrix` writes after running every lane locally
- pin the workflow's Fedora and Debian lanes to `dist/release-matrix.json` in `test/check-release-matrix.py`, so a new target cannot reach the matrix and miss the gate
- build the Fedora lanes' staged binaries in the pinned Arch container and declare them with `FACELOCK_RELEASE_BINARIES_PREBUILT=1`, checked rather than trusted
- keep the fixture's shadow entry reachable by pam_unix's setuid helper (`chmod 0400 /etc/shadow` in the rpm-e2e image), which is what the Fedora 43/44 lifecycle reds came down to
- dump the whole PAM stack when a lifecycle case fails: pamtester's output, daemon state, pam_unix's credential sources, the module's and helper's journal, and a replay under strace
- stage every Fedora containerfile the release checker reads, which `test/release-version-contract.sh` never did
- cover the classifier with `test/classify-changes-test.sh`, wired into `just check`

## Why

Every deb, rpm and Arch gate was local-only, invoked by hand, for a release that
ships through four channels. Nothing ran them between one maintainer remembering
to and a release tag firing `release.yml`, which is the worst place to find out.
Path filtering is what makes the cost bearable: the full matrix is 30-60+ minutes
and most pull requests touch no packaging.

## Files

| Path | Why it matters |
|---|---|
| `.github/workflows/packaging.yml` | the gate itself, and the three schedules *(start here)* |
| `.github/workflows/scripts/classify-changes.sh` | the path list, and the reasoning for hand-rolling it |
| `.github/actions/fetch-models/action.yml` | shared model fetch, now used by `container-pam-test` and every packaging lane |
| `justfile` | `test-packaging-matrix`, `test-classify-changes`, `_require-release-binaries`, and the preflight gate |
| `test/Containerfile.rpm-e2e` | the shadow-mode fixture accommodation that un-reds the Fedora lifecycle lanes, with the reasoning |
| `test/rpm-service-pam-lifecycle.sh` | the failure diagnostics that found it |
| `test/check-release-matrix.py` | lane-to-matrix drift check |
| `docs/releasing.md` | the residual risk, written down where a release is planned |

## Reviewer notes

- **models were the blocker, not the lanes.** Every package gate depends on `_require-models`, `models/*.onnx` are gitignored, and the only escape hatch downgrades the daemon-start assertions to skips. PR #280 already derived names, URLs and checksums from the manifest; this reuses that rather than writing a second copy, and caches the ~170MB on the manifest hash
- **no third-party filter action, and not GitHub's `paths:`.** #260 pinned every action by commit and took automerge away from them, so a filter action is another SHA to review every Monday. `paths:` skips the workflow, and a required check that never runs sits pending forever
- **running the gate for the first time found four real problems.** Ubuntu's `just` 1.21 cannot parse this justfile at all, so the lanes now install a checksum-pinned 1.46. `just build-release` cannot run on the runner: its TPM step wants tss2 4.1.3+ and Ubuntu 24.04 ships 4.0.1, so the Fedora lanes take binaries built in the Arch container instead. `just test-release-contract` was already failing everywhere, on a staging list that never grew as the checker learned to read more files. And the runner's container runtime cannot run Fedora's PAM helper against a mode-0000 shadow file (next note)
- **the Fedora 43/44 reds were the runner's container runtime, not the packages and not the module.** pam_unix 1.7+ verifies every shadowed password through an exec of its setuid `unix_chkpwd`, and on the runner's rootless podman that helper cannot see the entry behind Fedora's mode-0000 `/etc/shadow`, while a plain root shell in the same container reads it fine (getent rc=0, full capability set, NoNewPrivs=0, native overlay — the lifecycle diagnostics captured all of it). Every correct password then dies as `AUTHINFO_UNAVAIL`, read back as "correct password did not fall through". A module-free pam_unix-only stack fails identically on the runner, and the daemon answered the module's call with a correct no-match, so nothing of facelock's is involved; Debian never hits it because its shadow is 640 root:shadow. The fixture now sets the file 0400 — owner-read needs no capability at all — which keeps the real helper path working everywhere and changes nothing the lifecycle asserts
- **residual risk is deliberate and documented.** A Rust-only change that breaks the packaged runtime is not caught by its own pull request; nightly catches it within a day, the release gate before shipping. `docs/releasing.md` says so, so nobody reads a green PR as packaging-verified

## Validation

- `just check`, which now includes `test-classify-changes`
- `python3 test/check-release-matrix.py`, plus a deliberate lane-drift edit to confirm it fails closed
- `test/classify-changes-test.sh`, plus real branch diffs: docs-only classifies false, `debian/`-only classifies true
- `bash test/release-version-contract.sh` and `just test-release-native-ordering` locally
- `actionlint` and `shellcheck` on the new workflow and scripts
- the gate itself, on this pull request
- the Fedora 44 lifecycle lane locally against the run's own binary artifact, with the failure reproduced and the fix confirmed by deleting the fixture's shadow row and by the runner's own diagnostics

Closes #229

## Summary by CodeRabbit

- **New Features**
  - Added automated packaging validation across Debian, Fedora, and Arch environments.
  - Added scheduled and release-preflight packaging checks with change-aware execution.
  - Added reliable model retrieval with caching, retries, and checksum verification for CI tests.

- **Bug Fixes**
  - Improved PAM container testing and failure diagnostics.
  - Fixed packaging lane verification and release evidence requirements.

- **Documentation**
  - Updated testing, packaging, and release guidance to reflect the expanded CI coverage and preflight requirements.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -4,6 +4,7 @@ paths:
   - "justfile"
   - "crates/**/tests/**"
   - ".github/workflows/**"
+  - ".github/actions/**"
 ---
 
 # Testing Strategy
@@ -28,6 +29,34 @@ paths:
 Fedora recipes take a release and default to 44 (`just test-rpm-pkg 43`). Tier 3e
 covers all three declared targets at the depth `dist/release-matrix.json` gives
 each; Rawhide is experimental and never a lane.
+
+## What CI runs, and when
+
+`.github/workflows/ci.yml` gates every pull request: build, test, clippy, audit,
+the PAM standalone surface, agent docs, translation catalogs, and tier 3/3a in
+`container-pam-test`.
+
+`.github/workflows/packaging.yml` gates the packaged artifacts: tiers 3d and 3e,
+both Debian suite lanes, and the native version-ordering matrix. It runs on three
+schedules:
+
+| When | What | Filter |
+|---|---|---|
+| Pull request | every lane | only when the diff reaches a package |
+| Nightly (07:00 UTC) | every lane | none |
+| `just release-preflight` | evidence of a green run at HEAD | none |
+
+The pull-request filter is a `changes` job running
+`.github/workflows/scripts/classify-changes.sh`, plain bash over a merge-base
+diff. It is not GitHub's `paths:`, which strands a required check as pending
+forever, and not a third-party filter action, which would be another pinned SHA
+to review. Add a path there when a new file can reach a built package.
+
+So a green pull request is **not** packaging-verified unless the packaging jobs
+actually ran on it. A Rust-only change that breaks the packaged runtime is caught
+by the nightly matrix within a day, and by the release gate before it ships.
+`just test-packaging-matrix` runs every lane locally and records the commit, for
+a maintainer without CI in reach.
 
 ## Which E2E tier a new assertion belongs in
 
--- a/.claude/skills/packaging-test/SKILL.md
+++ b/.claude/skills/packaging-test/SKILL.md
@@ -5,13 +5,21 @@ description: Pick and run the right facelock packaging or container test for a c
 
 # Packaging and container tests
 
-CI runs exactly one container job, `container-pam-test` in `ci.yml`. It runs
-the PAM smoke tier and the camera-free E2E tier. Every `.deb`, `.rpm`, COPR and
-APT-repo path is **local only**. If you changed packaging and did not run one of
-these yourself, it is untested until a release tag fires `release.yml`, which is
-the worst place to find out.
+CI runs the packaging lanes in `.github/workflows/packaging.yml`: both Debian
+suite gates, every declared Fedora lane, the Arch package built from the real
+`dist/PKGBUILD`, and the native version-ordering matrix. On a pull request they
+run **only when the diff reaches a package**. A `changes` job classifies the
+merge-base diff, and every lane sits behind
+`if: needs.changes.outputs.packaging == 'true'`. Unfiltered runs happen nightly
+and at `just release-preflight`, which refuses to pass without a green matrix at
+HEAD.
 
-All recipes need `podman`. None of the ones in the routing table need a camera.
+Two consequences for you. A green pull request says nothing about packaging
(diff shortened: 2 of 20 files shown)


tyvsmith:
@codex review

PR #304, opened 2026-08-25
fix(security): wipe remaining decrypted embedding paths

### LINKED ISSUE #269: bench: cmd_calibrate leaves live-capture embeddings unwiped
`cmd_calibrate` captures face embeddings from live camera frames into a plain
`Vec<FaceEmbedding>` and drops it unwiped, in both bench implementations. The
stored templates loaded alongside them in the same function are wiped.

## Reproduction

1. `sudo facelock bench calibrate` (and the standalone `facelock-bench calibrate`)
2. `crates/facelock-cli/src/commands/bench.rs`, in `cmd_calibrate`: `enrolled` is
   wrapped in `Wiped`, while `live_embeddings` a few lines below is a bare
   `Vec<FaceEmbedding>` populated from captured frames and never zeroized.
3. Same shape in `crates/facelock-bench/src/main.rs` `cmd_calibrate`.

## Expected vs actual

Expected: every buffer holding face embeddings is zeroized on drop, including
unwind — the contract `authenticate_with_embeddings` documents and the one the
`Wiped` guard exists to enforce.

Actual: only the stored-template set is wiped. The live-capture set, which is a
fresh biometric read rather than a decrypted stored one, is left in process
memory.

## Scope

Both `cmd_calibrate` implementations only; the cold-auth, warm-auth and report
paths hold stored templates and are covered. Root-only developer commands, so no
unprivileged exposure. Surfaced while wiping the stored-template sites; those are
handled separately.

## Possible approach

`Wiped` is currently typed for `(u32, FaceEmbedding)` sets, so covering a bare
`Vec<FaceEmbedding>` needs a small generalisation of the guard rather than a
second implementation.

### LINKED ISSUE #293: decrypted embeddings drop unwiped on four error paths, one in the auth path
Four paths hold decrypted face embeddings and drop them unwiped. One is in the
authentication path, not a developer tool.

## Where

- `crates/facelock-daemon/src/embeddings.rs:88-120` — `decrypt_user_embeddings`
  hitting `?` mid-loop (AAD mismatch, TPM unseal failure on row N) drops rows
  1..N, already decrypted, unwiped. This is the shared load path under **auth**
  and preview.
- `crates/facelock-cli/src/direct.rs:187-194` — the set is loaded, then
  `store.list_models(user).context(...)?` can early-return before the wiping
  callee wraps it. The daemon avoids this by loading after `list_models` and
  hand-wiping its pre-callee error path (`handler.rs:1099-1104`); the direct
  path has no equivalent.
- `crates/facelock-daemon/src/auth.rs:553-559` — `.filter().cloned().collect()`
  has a zero lower size-hint, so the Vec grows by reallocation and frees
  intermediate buffers of decrypted embeddings unwiped.
  (`decrypt_user_embeddings` itself uses an exact `with_capacity`.)
- `crates/facelock-cli/src/commands/encrypt.rs:186-195` — `facelock tpm decrypt`
  holds unsealed template bytes per row and never wipes them. Note the `Wiped`
  guard is typed for embedding sets and cannot hold raw `Vec<u8>` blobs.

## Reproduction

Encrypted store with `bind_device_aad` and a replaced camera: row 3 fails its AAD
check, so rows 1-2 sit decrypted in freed daemon heap.

## Expected vs actual

Expected: every buffer holding decrypted templates is zeroized on drop, including
error and unwind paths — the contract `authenticate_with_embeddings` documents and
the `Wiped` guard exists to enforce.

Actual: the guard covers the sets its callers wrap; these four error paths release
plaintext before the wrap happens or outside its type.

## Scope

Pre-existing; not introduced by the preview/bench wiping work, which deliberately
scoped its docs claim to the compare sets it wraps. `direct.rs:187` is roughly a
two-line fix (reorder the load after `list_models`, or wrap at the load site);
`embeddings.rs` needs the partial result wiped before propagating the error.

Live-capture probe embeddings are a separate data class, tracked in #269.

tyvsmith:
## Summary

- generalize `Wiped` over its element (`WipeElement`: stored tuples, bare embeddings) and add `push`, which zeroizes the outgrown allocation instead of letting `Vec` free it live
- accumulate the decrypt loop inside the guard's borrowed form, so a row failing mid-load (AAD mismatch, TPM unseal) wipes the rows already decrypted
- build the auth compare set through the guard at exact capacity: `collect()` on the filtered iterator has a zero lower size-hint, so it grew by reallocation and freed live embeddings
- load the direct auth path's plaintext last, after `list_models`, matching the daemon's ordering
- hold `facelock tpm decrypt`'s per-row unsealed bytes in `zeroize::Zeroizing`; raw byte buffers already implement `Zeroize`, so `Wiped` gains no `u8` element
- guard both `calibrate` live-capture probe sets
- accumulate the store's plaintext fast load (`get_user_embeddings`) through the guard: it had the same zero-initial-capacity reallocation, plus an unwiped drop when a corrupt row fails the load mid-loop
- rescope the `docs/security.md` claim to the load and decrypt transients now covered, naming per-frame inference outputs, SQLite's per-row blob buffers, and `facelock tpm encrypt`'s raw-row snapshot as still unguarded

## Why

The guard covered the compare sets its callers wrapped; four error paths released plaintext before the wrap or outside its type, and one is the shared load path under authentication. With `bind_device_aad` and a replaced camera, a row failing its AAD check left every earlier row decrypted in freed daemon heap. The live-capture probes in `calibrate` were never wiped at all.

## Reviewer notes

- **`decrypt_rows_into` forgets its borrowed guard on success.** the guard owns only the borrow, so `mem::forget` skips the wipe without leaking; error and unwind exits still wipe the caller's buffer, which is what makes the new mid-loop tests observable
- **the guard's restrictions are unchanged.** still no `DerefMut`, `into_inner`, or `Clone`; `push` only adds

## Validation

- `just check`
- `cargo test --workspace`
- `just check-pam-standalone` and `cargo tree -p pam-facelock`
- reran the two mid-loop wipe tests with the guard removed to confirm they fail

Fixes #293
Fixes #269

## Summary by CodeRabbit

* **Security Enhancements**
  * Biometric embeddings and decrypted plaintext are now automatically zeroized after use, including during errors, panics, partial failures, and buffer growth.
  * Improved protection across calibration, authentication, comparison, database loading, and software- or TPM-backed decryption workflows.
  * Added safeguards for intermediate biometric data during collection and failed operations.

* **Documentation**
  * Expanded security documentation covering in-memory biometric data protection, cleanup scenarios, and remaining limitations.

### DIFF
--- a/crates/facelock-bench/src/main.rs
+++ b/crates/facelock-bench/src/main.rs
@@ -424,7 +424,11 @@ fn cmd_calibrate() -> Result<()> {
     let num_test_frames = 10u32;
     println!("Capturing {} test frames from camera...", num_test_frames);
 
-    let mut live_embeddings: Vec<FaceEmbedding> = Vec::new();
+    // Live-capture probe embeddings are biometric data like the stored set
+    // above: accumulated inside the guard, zeroized when it drops — bails,
+    // panics and buffer growth included (#269).
+    let mut live_embeddings: Wiped<Vec<FaceEmbedding>, FaceEmbedding> =
+        Wiped::with_capacity(num_test_frames as usize);
     for _ in 0..num_test_frames {
         let frame = camera.capture().context("Failed to capture frame")?;
         let faces = engine.process(&frame)?;
@@ -445,7 +449,7 @@ fn cmd_calibrate() -> Result<()> {
 
     // Compute all similarities between live and enrolled embeddings
     let mut similarities: Vec<f32> = Vec::new();
-    for live_emb in &live_embeddings {
+    for live_emb in live_embeddings.iter() {
         for (_, enrolled_emb) in enrolled.iter() {
             similarities.push(cosine_similarity(live_emb, enrolled_emb));
         }
--- a/crates/facelock-cli/src/commands/bench.rs
+++ b/crates/facelock-cli/src/commands/bench.rs
@@ -381,7 +381,11 @@ fn cmd_calibrate(config: &Config) -> Result<()> {
     let num_test_frames = 10u32;
     println!("Capturing {} test frames from camera...", num_test_frames);
 
-    let mut live_embeddings: Vec<FaceEmbedding> = Vec::new();
+    // Live-capture probe embeddings are biometric data like the stored set
+    // above: accumulated inside the guard, zeroized when it drops — bails,
+    // panics and buffer growth included (#269).
+    let mut live_embeddings: Wiped<Vec<FaceEmbedding>, FaceEmbedding> =
+        Wiped::with_capacity(num_test_frames as usize);
     for _ in 0..num_test_frames {
         let frame = camera.capture().context("Failed to capture frame")?;
         let faces = engine.process(&frame)?;
@@ -402,7 +406,7 @@ fn cmd_calibrate(config: &Config) -> Result<()> {
 
     // Compute all similarities between live and enrolled embeddings
     let mut similarities: Vec<f32> = Vec::new();
-    for live_emb in &live_embeddings {
+    for live_emb in live_embeddings.iter() {
         for (_, enrolled_emb) in enrolled.iter() {
             similarities.push(cosine_similarity(live_emb, enrolled_emb));
         }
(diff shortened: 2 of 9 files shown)


tyvsmith:
@codex review

PR #372, opened 2026-09-10
test(packaging): Probe the deb-only lane selection

tyvsmith:
throwaway probe for #371: a deb-harness-only diff should run the two Debian lanes and the ordering matrix, and skip arch, release binaries and Fedora. closed unmerged once observed


tyvsmith:
observed: deb trixie/resolute + ordering matrix ran; arch, release binaries and Fedora skipped (not pending). run cancelled to save minutes

PR #383, opened 2026-09-12
test(release): Roll the upgrade predecessor with the matrix, not a literal

### LINKED ISSUE #367: release matrix still pins v0.1.4 as the predecessor, so 0.2.0 to 0.2.1 is never exercised
The release matrix pins v0.1.4 as the released predecessor, so the upgrade lanes prove 0.1.4 to candidate and preflight compares COPR against 0.1.4-1. The release users will actually perform after 0.2.1 ships is 0.2.0 to 0.2.1, and nothing in CI exercises it.

## Reproduction

- `dist/release-matrix.json` `predecessors` holds only `v0.1.4`
- `test/upgrade-v014-predecessor.sh:49` reads `matrix["predecessors"]["v0.1.4"]` by literal key
- `just release-preflight v0.2.1` failed on 2026-09-07 because COPR serves 0.2.0-1, which matches neither the pin nor the recorded 0.1.3-1 gap; #366 re-records the gap so 0.2.1 can ship

## Expected vs actual

Expected: after a release ships, the pin rolls to it, the gap record retires, and the deb and rpm upgrade lanes prove predecessor-to-candidate from the newest released assets.

Actual: the pin, the lane script's key, the lane names (`upgrade-v014-*`), and v0.1.4-specific accommodations in the lane images (hand-written control file, undeclared libxkbcommon0, V5 to V6 assertions) all assume v0.1.4.

## Scope

- roll `predecessors` to v0.2.0: release id, asset ids, sha256 and size for the trixie deb and the fc44 rpm, resolved from `gh api repos/tyvsmith/facelock/releases/tags/v0.2.0`
- delete `served_evr_gap`; `docs/releasing.md:756` already says it retires when the pin moves
- have the lane script and images read the pinned tag rather than a literal; keep v0.1.4 accommodations only where the predecessor still needs them
- consider pinning two predecessors (the last release and the last minor) so both upgrade paths stay proven; `check-live-release-channels.py` currently requires exactly one
- relates to #231 (upgrade preservation proof) and #333 (COPR served-EVR gap)

tyvsmith:
## Summary

- name the release the upgrade lanes prove in `dist/release-matrix.json` as `predecessors.current`, pinned to v0.2.1's .deb and fc44 .rpm by asset id, digest and size
- require `current` to be the newest pinned release, so a pin left behind fails `test/check-release-matrix.py`
- keep v0.1.4 pinned: the retired-authselect fixture is built from its RPM and no other build of that package exists
- resolve the predecessor through `current` in the lane pin resolver, the candidate version, and `check-live-release-channels.py --expect-predecessor` (what `just release-preflight` asks)
- delete `copr_channels.production.served_evr_gap`; production COPR serves the EVR the new pin owes
- key the lane's release-specific behaviour off the pinned version rather than a tag: the ADR 009 `tpm encrypt` spelling, the ADR 010 layout, the pre-0.2.0 `libxkbcommon0` accommodation
- record the schema the predecessor writes instead of asserting V5, and hold a V7 predecessor's `key_id` across the upgrade
- create the seed database by starting the released daemon, since `tpm encrypt` refuses a database that is not there from 0.2.1 on, and plant the stale enrollment marker after that so nothing reconciles it before the upgrade does
- fail the lane when the candidate daemon never opens the database, which the schema assertion alone stopped proving once the predecessor already wrote V7
- default the candidate's upgrade-test version to the predecessor's patch plus one
- rename `just test-upgrade-v014*` to `just test-upgrade-predecessor*`, and the files behind them

## Why

The lanes and the preflight served-EVR gate both read the predecessor by the literal key `v0.1.4`, so they kept proving 0.1.4-to-candidate through 0.2.0 and 0.2.1 while every user was upgrading from one of those. Preflight failed on 2026-09-07 for the same reason: it asked production COPR for 0.1.4-1 while the channel served 0.2.0-1, and #366 papered over that with a gap record rather than moving the pin. Making the pin a matrix field means rolling it after a release is a data edit, and the contract now fails when it is skipped.

## Validation

- `just test-upgrade-predecessor-contract`, `test/release-version-contract.sh`, `python3 test/check-release-matrix.py`, `just test-release-artifacts`, `just check-docs`
- `just test-upgrade-predecessor-pins` against the live release API
- `check-live-release-channels.py --expect-predecessor` against live production COPR
- the Debian upgrade lane locally against the v0.2.1 predecessor: all five state shapes, rollback, and the fault matrix

The Fedora half is unrun here; both lanes are local-only by design and no CI job carries them.

Fixes #367

## Summary by CodeRabbit

- **Changed**
  - Upgrade testing now targets the newest released predecessor selected from the release matrix, rather than a fixed version.
  - Upgrade and rollback coverage now supports varying predecessor versions, database schemas, package layouts, and encryption behavior.
  - Release validation now confirms predecessor pins, package metadata, and served release-channel versions remain consistent.
  - Upgrade-test commands and related documentation now use the `test-upgrade-predecessor` naming.

- **Documentation**
  - Updated testing, release, contract, and developer guidance to describe predecessor-driven upgrade validation.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -19,7 +19,7 @@ paths:
 | 3d | Arch package from `dist/PKGBUILD` | `just test-arch-pkg` |
 | 3e | Fedora package lifecycle, every declared release | `just test-rpm-lanes` |
 | 3f | Fedora COPR path, rebuilt from source on system ORT | `just test-copr-lanes` |
-| 3g | Released v0.1.4 upgrade and rollback, deb and rpm | `just test-upgrade-v014` |
+| 3g | Released-predecessor upgrade and rollback, deb and rpm | `just test-upgrade-predecessor` |
 | 3a | Arch container E2E, camera-free | `just test-arch-camera-free` |
 | 3b | Arch container E2E (daemon), needs a camera | `just test-arch-integration` |
 | 3c | Arch container E2E (oneshot), needs a camera | `just test-arch-oneshot` |
--- a/.claude/skills/packaging-test/SKILL.md
+++ b/.claude/skills/packaging-test/SKILL.md
@@ -44,7 +44,7 @@ need `podman`; none in the routing table needs a camera.
 | `crates/pam-facelock/**`, `/etc/pam.d` handling | `just test-arch-pam` and `just check-pam-standalone` |
 | File layout, installed paths | `just test-arch-layout` |
 | D-Bus policy, daemon authorization, pre-flight exit codes, schema migrations | `just test-arch-camera-free` |
-| Store migrations, encryption keys, TPM sealing, maintainer scripts touching state | `just test-upgrade-v014` |
+| Store migrations, encryption keys, TPM sealing, maintainer scripts touching state | `just test-upgrade-predecessor` |
 
 `just test-deb` delegates to both exact supported-suite package gates. The
 remaining quick syntax-level check is `just test-rpm` (Fedora container), which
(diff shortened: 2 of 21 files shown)


tyvsmith:
@coderabbitai review

